Issue #7, opened 2020-07-27
lagoon-logging: end2end tests and ansible service broker create high amount of indexes


Schnitzel:
on our openshift clusters we have an end2end test, which tests the creation of a new project, deploys a pod and removes the namespace again. This runs every 15min and creates 24*4 namespaces every day, therefore 96 indexes for the container logs:

```
container-logs-e2e-1594921999-4f5dk34jlba7lmvmjnc2hslnum_bi-_-unknown_environmenttype-_-2020.07 
container-logs-e2e-1595660042-m45wplgchbg5jgioiabnaf36ya_bi-_-unknown_environmenttype-_-2020.07
container-logs-e2e-1595105636-vggrvpciebffvefnrowsuquz3q_bi-_-unknown_environmenttype-_-2020.07
```

the same is for the ansible servicebroker which creates new namespaces as well:

```
container-logs-lagoon-dbaas-mariadb-apb-prov-nqr5n_bi-_-unknown_environmenttype-_-2020.07
container-logs-lagoon-dbaas-mariadb-apb-prov-gsdxw_bi-_-unknown_environmenttype-_-2020.07
container-logs-lagoon-dbaas-mariadb-apb-depr-gsdxw_bi-_-unknown_environmenttype-_-2020.07
```

in lagoon we captured them here:
https://github.com/amazeeio/lagoon/blob/8128a4db750a61fcda62e6d5206be6c4865506e9/services/logs2logs-db/pipeline/router-application-logs.conf#L68-L72
https://github.com/amazeeio/lagoon/blob/8128a4db750a61fcda62e6d5206be6c4865506e9/services/logs-forwarder/.lagoon.single.yml#L194-L211

can we implement something similar in lagoon-logging again? It keeps our elasticsearch cluster small.

I'm aware that this is maybe just a special think for amazee.io related cluster and might not be something that would be good for every lagoon installation, maybe we can do a similar thing to `excludeNamespaces` from the loggingOperator also for the logs-dispatcher?

smlx:
hi @Schnitzel I've pushed a PR that fixes this issue, but it's hard-coding those particular indices. Are you okay with that solution or do you think we really need the flexibility of making this configurable without doing a new chart version release?

Schnitzel:
@smlx 
good question, I guess I more thought about a way to ignore specific namespaces all together, instead of consolidating them. but I your code is not as intrusive that I thought, so I think we can just ship with that one and close this :)

smlx:
[...]

PR #9, opened 2020-07-28
Consolidate service indices

smlx:
This PR consolidates log indices coming from internal cluster services.

Closes: #7

### DIFF
--- a/charts/lagoon-logging/Chart.yaml
+++ b/charts/lagoon-logging/Chart.yaml
@@ -18,7 +18,7 @@ type: application
 # time you make changes to the chart and its templates, including the app
 # version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.6.5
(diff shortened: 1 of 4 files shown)


Schnitzel:
actually I think we need to match a bit wider for the second regex, as some indexes could be called like this:

```
container-logs-appuio-mariadb-apb-prov-nqr5n_bi-_-unknown_environmenttype-_-2020.07 
container-logs-foo-bar-apb-prov-nqr5n_bi-_-unknown_environmenttype-_-2020.07 
```

basically it's this way:

`container-logs-[ansible-name]-apb-[prov/depr]`

where the `ansible-name` is the name of the service broker. Most of the time this is called `lagoon-dbaas-mariadb` but we also have some others running in other clusters.

smlx:
@Schnitzel I made this configurable. If it looks okay please merge.

PR #10, opened 2020-07-29
Flush performance

Schnitzel:
this improves the flush performance of the whole logging system drastically, in my tests I had a container-log from a remote cluster within 20secs in kibana, while before this it took 3-4 minutes.

### DIFF
--- a/charts/lagoon-logging/templates/logs-dispatcher.fluent-conf.configmap.yaml
+++ b/charts/lagoon-logging/templates/logs-dispatcher.fluent-conf.configmap.yaml
@@ -328,6 +328,8 @@ data:
         tls_verify_hostname {{ . }}
         {{- end }}
         # endpoint
+        keepalive true # makes sure the connection is not recreated every second
+        keepalive_timeout 10m # reconnect after 10mins in order to handle DNS changes, etc.
         <server>
           port "#{ENV['LOGS_FORWARD_HOST_PORT']}"
           host "#{ENV['LOGS_FORWARD_HOST']}"
--- a/charts/lagoon-logging/templates/logs-dispatcher.fluent-conf.configmap.yaml
+++ b/charts/lagoon-logging/templates/logs-dispatcher.fluent-conf.configmap.yaml
@@ -352,7 +352,7 @@ data:
           flush_thread_count 8 # 8 threads that flush, makes sure we have enough flushers for all the different buffers
           flush_interval 2s # flush every 2 seconds
           flush_thread_burst_interval 0 # don't sleep if there is more data to flush
-          retry_type periodic # no exponential_backoff as this could lead to fluentd wait for easily 2h until it retries
+          retry_max_interval 30s # limit exponential backoff period
           overflow_action drop_oldest_chunk
         </buffer>
       </store>
--- a/charts/lagoon-logging/templates/logs-dispatcher.fluent-conf.configmap.yaml
+++ b/charts/lagoon-logging/templates/logs-dispatcher.fluent-conf.configmap.yaml
@@ -349,7 +349,10 @@ data:
           # buffer params (per worker)
           total_limit_size 8GB
           # flush params
-          flush_thread_count 2
+          flush_thread_count 8 # 8 threads that flush, makes sure we have enough flushers for all the different buffers
+          flush_interval 2s # flush every 2 seconds
+          flush_thread_burst_interval 0 # don't sleep if there is more data to flush
+          retry_type periodic # no exponential_backoff as this could lead to fluentd wait for easily 2h until it retries
           overflow_action drop_oldest_chunk
         </buffer>
       </store>
--- a/charts/lagoon-logging/values.yaml
+++ b/charts/lagoon-logging/values.yaml
@@ -233,12 +233,18 @@ excludeNamespaces:
 # This may require tweaking to handle high volumes of logs.
 clusterOutputBuffer:
   flush_thread_count: 32
-  timekey: 10s # create chunks every 10 secs, therefore possible to flush every 10sec
-  timekey_wait: 0s # don't wait additional time for the timekey to be filled
+  # create chunks every 10 secs, therefore possible to flush every 10sec
+  timekey: 10s
+  # don't wait additional time for the timekey to be filled
+  timekey_wait: 0s
   timekey_use_utc: true
-  flush_thread_burst_interval: '0' # don't sleep if there is more data to flush
-  flush_mode: immediate # flush immediate, the forward target is in the same cluster so this should be able to handle it
-  retry_max_interval: 30s # limit exponential backoff period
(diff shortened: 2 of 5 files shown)


charts/lagoon-logging/values.yaml
```diff
@@ -233,9 +233,12 @@ excludeNamespaces:
 # This may require tweaking to handle high volumes of logs.
 clusterOutputBuffer:
   flush_thread_count: 32
-  timekey: 1m
-  timekey_wait: 10s
+  timekey: 10s # create chunks every 10 secs, therefore possible to flush every 10sec
+  timekey_wait: 0s # don't wait additional time for the timekey to be filled
   timekey_use_utc: true
+  flush_thread_burst_interval: '0' # don't sleep if there is more data to flush
+  flush_mode: immediate # flush immediate, the forward target is in the same cluster so this should be able to handle it
```
smlx, line 240: I did do quite a bit of testing around this configuration and was able to overload the target fluentd fairly easily. I think we could remove this line if it's going to flush in <10s anyway.

charts/lagoon-logging/templates/logs-dispatcher.fluent-conf.configmap.yaml
```diff
@@ -299,6 +299,8 @@ data:
         tls_verify_hostname {{ . }}
         {{- end }}
         # endpoint
+        keepalive true # makes sure the connection is not recreated every second
+        keepalive_timeout 10m # reconnect after 10mins in order to handle DNS changes, etc.
```
smlx, line 303: The problem with this is that it breaks load-balancing on the target forwarder because a standard kubernetes load balancer does round-robin on the targeted endpoints. So e.g. with this setting logs will flow to a single pod, switching pods every 10m.

charts/lagoon-logging/templates/logs-dispatcher.fluent-conf.configmap.yaml
```diff
@@ -318,7 +320,10 @@ data:
           # buffer params (per worker)
           total_limit_size 8GB
           # flush params
-          flush_thread_count 2
+          flush_thread_count 8 # 8 threads that flush, makes sure we have enough flushers for all the different buffers
+          flush_interval 2s # flush every 2 seconds
```
smlx, line 324: Without the keepalive 2s might be too short for full TCP handshake and teardown. I wonder if we could bump this to 10s or more?

charts/lagoon-logging/templates/logs-dispatcher.fluent-conf.configmap.yaml
```diff
@@ -318,7 +320,10 @@ data:
...
+          flush_thread_burst_interval 0 # don't sleep if there is more data to flush
+          retry_type periodic # no exponential_backoff as this could lead to fluentd wait for easily 2h until it retries
```
smlx, line 326: Ah, good point about the 2h wait. `retry_max_interval 60s` is another option here - I don't know which is better :shrug:

charts/lagoon-logging/values.yaml
```diff
@@ -233,9 +233,12 @@ excludeNamespaces:
 # This may require tweaking to handle high volumes of logs.
 clusterOutputBuffer:
   flush_thread_count: 32
-  timekey: 1m
-  timekey_wait: 10s
+  timekey: 10s # create chunks every 10 secs, therefore possible to flush every 10sec
```
smlx, line 236: Due to the volume of logs coming from the cluster I actually ran into issues with setting the timekey too low. I was getting errors like this due to running out of file descriptors due to lots of little chunks:
```
<Fluent::Config::Section {"@label":"@d18536be251369c8ce5b2628a6f329f2","tag":"","matches":["<Fluent::Config::Section {\"labels\":{},\"namespaces\":[],\"hosts\":[],\"container_names\":[],\"negate\":false}>"]}>
#<Thread:0x000056346293a188@event_loop@/usr/lib/ruby/gems/2.6.0/gems/fluentd-1.9.2/lib/fluent/plugin_helper/thread.rb:70 run> terminated with exception (report_on_exception is true):
/usr/lib/ruby/2.6.0/socket.rb:1313:in `__accept_nonblock': No file descriptors available - accept(2) (Errno::EMFILE)
	from /usr/lib/ruby/2.6.0/socket.rb:1313:in `accept_nonblock'
	from /usr/lib/ruby/gems/2.6.0/gems/cool.io-1.6.0/lib/cool.io/listener.rb:46:in `on_readable'
	from /usr/lib/ruby/gems/2.6.0/gems/cool.io-1.6.0/lib/cool.io/loop.rb:88:in `run_once'
	from /usr/lib/ruby/gems/2.6.0/gems/cool.io-1.6.0/lib/cool.io/loop.rb:88:in `run'
	from /usr/lib/ruby/gems/2.6.0/gems/fluentd-1.9.2/lib/fluent/plugin_helper/event_loop.rb:93:in `block in start'
	from /usr/lib/ruby/gems/2.6.0/gems/fluentd-1.9.2/lib/fluent/plugin_helper/thread.rb:78:in `block in thread_create'
```
Have you seen any errors like that in the `logging-operator` fluentd?

charts/lagoon-logging/templates/logs-dispatcher.fluent-conf.configmap.yaml
smlx: As discussed, maybe we could make this 1m to avoid loading a single pod for too long?

charts/lagoon-logging/values.yaml
smlx: As discussed, lets keep this since logs-dispatcher seems to handle it okay.

charts/lagoon-logging/values.yaml
smlx: Checked the cluster and don't see any issues. If there are problems we can always raise this again :+1:

charts/lagoon-logging/values.yaml
```diff
@@ -233,9 +233,12 @@ excludeNamespaces:
...
+  timekey_wait: 0s # don't wait additional time for the timekey to be filled
   timekey_use_utc: true
+  flush_thread_burst_interval: '0' # don't sleep if there is more data to flush
```
smlx, line 239: Should this be an integer instead of a string?

smlx: https://github.com/banzaicloud/logging-operator/blob/master/pkg/sdk/model/output/buffer.go#L72
Apparrently not

charts/lagoon-logging/templates/logs-dispatcher.fluent-conf.configmap.yaml
smlx: Will leave it, can always tweak should issues arise.

smlx (approved):

PR #18, opened 2020-08-05
Lagoon core chart

smlx:
Initial chart for lagoon-core:
* api
* api-db
* api-redis
* keycloak
* keycloak-db
* broker

Closes: amazeeio/lagoon#2007
Closes: amazeeio/lagoon#2008
Closes: amazeeio/lagoon#2009
Closes: amazeeio/lagoon#2010
Partially addresses: amazeeio/lagoon#1597


tobybellwood:
ah, helmpack/chart-testing is only using helm 3.1.2, which has a linting fail for the lookup function that was fixed in 3.2.0
https://github.com/helm/chart-testing/blob/master/Dockerfile#L31
https://github.com/helm/helm/security/advisories/GHSA-q8q8-93cv-v6h8

PR #19, opened 2020-08-05
First lagoon-ssh chart draft

vincenzodnp:
This is the first draft of lagoon-ssh chart.

Closes: amazeeio/lagoon#2013


charts/ssh/templates/ssh.deployment.yaml
```diff
@@ -0,0 +1,77 @@
...
+    matchLabels:
+      {{- include "ssh.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      labels:
+        {{- include "ssh.selectorLabels" . | nindent 8 }}
+    spec:
+      containers:
+        - name: {{ .Chart.Name }}
+          env:
```
smlx, line 19: You can just use the entire secret as an env here: https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/#configure-all-key-value-pairs-in-a-secret-as-container-environment-variables

charts/ssh/templates/ssh.deployment.yaml
```diff
@@ -0,0 +1,77 @@
...
+            valueFrom:
+              secretKeyRef:
+                key: RABBITMQ_PASSWORD
+                name: ssh-env
+          - name: JWTSECRET
+            valueFrom:
+              secretKeyRef:
+                key: JWTSECRET
+                name: ssh-env
+          - name: SERVICE_NAME
```
smlx, line 35: I don't know if this variable and `CRONJOBS` are used - if not, we can just remove them.

charts/ssh/templates/ssh.deployment.yaml
```diff
@@ -0,0 +1,77 @@
...
+                key: API_DB_PASSWORD
+                name: ssh-env
+          securityContext: {}
+          image: "{{ .Values.ssh.image.repository }}:{{ .Values.ssh.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.ssh.image.pullPolicy }}
+          ports:
+            - name: ssh
+              containerPort: {{ .Values.ssh.containerPort }}
+              protocol: TCP
+          livenessProbe:
```
smlx, line 50: You probably don't need to set all these values, because the defaults are pretty sensible. e.g. this might be enough:
```
livenessProbe:
  tcpSocket:
    port: ssh
```
Also some of these settings are set to default values. e.g. `failureThreshold` defaults to 3. see [here](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes) for the defaults.

Same for the `readinessProbe`.

PR #60, opened 2020-09-08
Update logo

smlx:
Closes: #53


tobybellwood:
I'd replace the IO logo with a Lagoon one?
lagoon

smlx:
That's the one I used? https://github.com/uselagoon/lagoon-charts/pull/60/files#diff-cc35d4b8e796c869289176c383aa3da3

tobybellwood:
ah yes - no idea why I was getting the old IO one...

happy with either - I've been using the one with the text more often, unless the scale is too small (favicon etc)

smlx:
Well I was thinking that the main place this will be seen is on helm hub and they use a square frame for the logo, so went with the square textless one. We can always replace it later if you like :slightly_smiling_face:

PR #88, opened 2020-09-23
remove registry

Schnitzel:
I don't think this is really needed and a leftover from old OpenShift times


tobybellwood (approved):

tobybellwood:
Registry is used at https://github.com/amazeeio/lagoon/blob/main/images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh#L959-L960 to store ?harbor?

Schnitzel:
@tobybellwood 
true, but it's actually set here: https://github.com/amazeeio/lagoon/blob/master/images/kubectl-build-deploy-dind/build-deploy.sh#L62 so I don't think the env variable itself is used.

PR #90, opened 2020-09-23
self discovery of uiURL and webhookURL

Schnitzel:


charts/lagoon-core/templates/api.deployment.yaml
```diff
@@ -122,9 +122,17 @@ spec:
           value: {{ . | quote }}
         {{- end }}
         - name: UI_URL
-          value: {{ required "A valid .Values.uiURL required!" .Values.uiURL | quote }}
+          {{- if .Values.lagoonUIURL }}
```
tobybellwood, line 125: why do we change the key name here from uiURL to lagoonUIURL?

tobybellwood (approved):

charts/lagoon-core/templates/api.deployment.yaml
Schnitzel: because we use already `lagoonAPIURL` in other places, so it's more clear now, things which are done by Lagoon are prefixed by `lagoon` :)

PR #100, opened 2020-10-01
broker: ingress and external amqp

Schnitzel:
this adds:
- ingress to see the rabbitmq mgmt interface
- external amqp loadbalancer as the deploy/build controller needs access to rabbitmq directly


charts/lagoon-core/templates/broker.service.yaml
```diff
@@ -40,3 +40,20 @@ spec:
...
+    {{- include "lagoon-core.broker.labels" . | nindent 4 }}
+spec:
+  type: {{ .Values.broker.service.amqpExternal.type }}
+  ports:
+  - port: {{ .Values.broker.service.amqpExternal.port }}
+    targetPort: amqp
+    name: amqp
+  selector:
+    {{- include "lagoon-core.broker.selectorLabels" . | nindent 4 }}
+{{- end }}
```
smlx, line 59: Missing a newline here.

smlx (changes requested):
Is the the broker web UI only being used for debugging? If so, I've been relying on:
```
kubectl port-forward svc/lagoon-core-broker http
```

Will the broker UI ever be used by someone who can't port-forward, or is there another reason you would want to expose an ingress?

charts/lagoon-core/templates/broker.service.yaml
Schnitzel: fixed :)

Schnitzel:
> Will the broker UI ever be used by someone who can't port-forward, or is there another reason you would want to expose an ingress?

@smlx good question, I could imagine that we have people that help with running lagoon and troubleshooting problems on lagoon that would not have kubectl access to the cluster. The same case is also happening at Harbor: technically nobody except lagoon admins need access to the Harbor UI but we still deploy it.

smlx (approved):
LGTM

PR #101, opened 2020-10-01
add API_HOST back to api, load harbor password from secret

Schnitzel:
- realized that the api actually needs API_HOST after all
- harbor password is stored in a secret so we can load it from there.


smlx:
Hey @Schnitzel I added `API_HOST` back already in https://github.com/uselagoon/lagoon-charts/pull/98

Schnitzel:
> Hey @Schnitzel I added `API_HOST` back already in #98

yep, sorry saw your PR after I pushed this already, will update this

smlx (approved):
LGTM

We probably shouldn't be generating this value in the secret though, because it is configured in the external harbor service. I need to go back and review those generated values.

smlx (approved):

PR #103, opened 2020-10-01
Add Backwards Compatibility Support for Harbor v1 API

cdchris12:
This PR adds the `HarborAPIVersion` value, which allows us to maintain backwards compatibility with Harbor v1 while also supporting Harbor v2 as well. This PR works in tandem with the variable introduced in [Lagoon PR #2231](https://github.com/amazeeio/lagoon/issues/2231) to provide this functionality.

### DIFF
--- a/charts/lagoon-core/templates/api.deployment.yaml
+++ b/charts/lagoon-core/templates/api.deployment.yaml
@@ -71,6 +71,8 @@ spec:
             secretKeyRef:
               name: {{ include "lagoon-core.api.fullname" . }}
               key: HARBOR_ADMIN_PASSWORD
+        - name: HARBOR_API_VERSION
+          value: {{ required "A valid .Values.harborAPIVersion required!" .Values.harborURL | quote }}
         - name: HARBOR_URL
           value: {{ required "A valid .Values.harborURL required!" .Values.harborURL | quote }}
         - name: JWTSECRET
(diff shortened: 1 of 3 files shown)


charts/lagoon-core/values.yaml
```diff
@@ -3,6 +3,7 @@
 # elasticsearchHost:
 # harborURL:
 # harborAdminPassword:
+# harborAPIVersion:
```
Schnitzel, line 6: can we have a default on harbor api v2? and then just if somebody really wants to use v1 they can change the apiversion?

Schnitzel (changes requested):

charts/lagoon-core/values.yaml
cdchris12: Great idea. I committed that change.

charts/lagoon-core/templates/api.deployment.yaml
```diff
@@ -70,6 +70,8 @@ spec:
         - name: HARBOR_ADMIN_PASSWORD
           value: {{ . | quote }}
         {{- end }}
+        - name: HARBOR_API_VERSION
+          value: {{ .Values.harborAPIVersion | default "v2.0" | quote }}
```
smlx, line 74: This works, but what we have done for other default values is to set them in `values.yaml` for the chart instead of putting the value itself in the template. Could we do it that way instead?

For example see the `rabbitMQUsername`, `elasticsearchScheme`, `elasticsearchHostPort` values.

smlx (changes requested):

charts/lagoon-core/templates/api.deployment.yaml
Schnitzel: +1 to this

smlx:
sorry for the delay on this!

PR #133, opened 2020-10-25
Lagoon Monitoring

Schnitzel:
This adds Lagoon Monitoring:
1. a helmfile for a Prometheus stack monitoring Lagoon components (elasticsearch, fluentd, logging-operator, broker, etc.)
2. ServiceMonitors for Lagoon Components so it can be picked up by LagoonMonitoring


charts/lagoon-core/templates/broker.servicemonitor.yaml
```diff
@@ -0,0 +1,17 @@
...
+metadata:
+  name: {{ include "lagoon-core.broker.fullname" . }}
+  labels:
+    {{- include "lagoon-core.broker.labels" . | nindent 4 }}
+spec:
+  endpoints:
+    - port: metrics
+  namespaceSelector:
+    matchNames:
+      - lagoon-core
```
smlx, line 13: I think this should be something like:
```suggestion
      - {{ .Release.Namespace }}
```

charts/lagoon-logging/templates/logs-dispatcher.servicemonitor.yaml
```diff
@@ -0,0 +1,21 @@
+{{- if .Values.logsDispatcher.serviceMonitor.enabled }}
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: {{ include "lagoon-logging.logsDispatcher.fullname" . }}
+  labels:
+    {{- include "lagoon-logging.logsDispatcher.labels" . | nindent 4 }}
+    monitoring.lagoon.sh/monitorMe: 'true'
```
smlx, line 8: Interesting, is this currently used?

helmfiles/lagoon-monitoring/auth.yaml.gotmpl
```diff
@@ -0,0 +1,14 @@
...
+  paths: ["/"]
+  hosts:
+    - {{ .Values.auth.host }}
+  tls:
+    - secretName: {{ .Values.auth.host }}-tls
+      hosts:
+        - {{ .Values.auth.host }}
+
+config:
+{{- toYaml .Values.auth.vouchConfig | nindent 2 }}
```
smlx, line 14: Missing a newline here.

helmfiles/lagoon-monitoring/values.yaml
```diff
@@ -0,0 +1,26 @@
...
+auth:
+  enabled: false
+  host: auth.monitoring.example.com
+  config:
+    vouch:
+      domains: []
+      allowAllUsers: false
+      whiteList: []
+      jwt:
+        secret: super-secret-stuff
```
smlx, line 11: So is the idea that this file is just example values that any user of the `helmfile.yaml` would need to override? If so maybe we could add a comment at the top?

smlx (changes requested):
A couple of minor comments, but generally looks great! :tada:

charts/lagoon-logging/templates/logs-dispatcher.servicemonitor.yaml
Schnitzel: @smlx you mean the `monitoring.lagoon.sh/monitorMe` ? Yes it's used by lagoon-monitoring prometheus
https://github.com/uselagoon/lagoon-charts/pull/133/files#diff-31c0db3ddbe675f8065e5156d8db158ca81b36a321026a4d0488df167841688aR259

helmfiles/lagoon-monitoring/values.yaml
Schnitzel: ah yes, will add a comment :)

smlx (approved):
LGTM once CI passes

maybe we should squash + merge since there are a lot of commits here?

Schnitzel:
> maybe we should squash + merge since there are a lot of commits here?
haha, yes definitely :)

PR #142, opened 2020-10-29
Safe underscore replacement

smlx:
There are a couple of issues with the change from https://github.com/uselagoon/lagoon-charts/pull/140

1. the result of `record.dig()` may be nil, so calling `sub()` will raise an exception (and cause the worker thread to restart). This PR fixes that by using the lonely operator.
2. `sub()` only replaces the _first_ match. This PR changes it to use `gsub()` instead.

The result of this logic is:

* the project part of the index names in Elasticsearch will have their underscores replaced by hyphens. 
* the field `lagoon_sh/project` within the log record will still have the original project name with any underscores.

Is that the way it should be @Schnitzel?

### DIFF
--- a/charts/lagoon-logging/Chart.yaml
+++ b/charts/lagoon-logging/Chart.yaml
@@ -18,7 +18,7 @@ type: application
 # time you make changes to the chart and its templates, including the app
 # version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.14.2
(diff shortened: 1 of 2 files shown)


smlx:
I found https://github.com/amazeeio/lagoon/blob/main/services/api/src/resources/group/opendistroSecurity.ts#L18-L22 which seems to confirm that this logic is correct.

PR #144, opened 2020-11-03
Build Deploy controller Openshift and Build Image support

shreddedbacon:
The Lagoon Build Deploy controller v0.1.4 supports Openshift now, this PR adds some notes in the `values.yaml` file with information around this. Also updated the CRD to the latest version.

Quick run down on the changes:

The controller supports defining the build image that gets injected into a Lagoon Build, as this will no longer be handled by Lagoon core in the future. 

Currently Lagoon will inject the `image` into the build spec and send that to the controller, and the controller will use what is in the spec before it uses what the controller has defined. This is on purpose, as there may be changes in Lagoon in the future that could allow for custom build images to be injected directly into the build per project/environment and override what is defined by the controller. 

The controller will default to using `uselagoon/kubectl-build-deploy-dind:latest` as the image if one is not defined anywhere else.

I think this can be merged any time as the existing functionality of the controller is unchanged in v0.1.4 of the controller.

This is work that was done to support Openshift using the controller for Lagoon https://github.com/amazeeio/lagoon/pull/2216

### DIFF
--- a/charts/lagoon-remote/Chart.yaml
+++ b/charts/lagoon-remote/Chart.yaml
@@ -18,7 +18,7 @@ type: application
 # This is the chart version. This version number should be incremented each
 # time you make changes to the chart and its templates, including the app
 # version.
-version: 0.5.0
+version: 0.5.1
 
 # This is the version number of the application being deployed. This version
 # number should be incremented each time you make changes to the application.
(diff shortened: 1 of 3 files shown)


smlx (approved):
sorry this took a while to get to. looks great :rocket:

PR #152, opened 2020-11-10
Reference correct generation Lagoon images and support .Values.imageTag

tobybellwood:
This PR references the `testlagoon/` dockerhub organisation - and sets the `:main` branch for images.

It also adds the necessary logic into all the charts to allow for a top-level .Values.imageTag to be set to override the default image tag behaviour.

Once we have a tagged release of `uselagoon/`-based images, we should switch to them, and then override them with testlagoon images via an additional values.yaml file or similar.


smlx:
@tobybellwood as we are using these charts already outside testing I'm going to rework this a bit to put the updated registry location only in the CI values files.

PR #156, opened 2020-11-14
consolidate e2e for router logs

Schnitzel:
the current e2e consolidation only matches containers (see the filter tag). This makes sure it also matches router logs

### DIFF
--- a/charts/lagoon-logging/templates/logs-dispatcher.fluent-conf.configmap.yaml
+++ b/charts/lagoon-logging/templates/logs-dispatcher.fluent-conf.configmap.yaml
@@ -175,7 +175,7 @@ data:
       @type record_modifier
       <replace>
         key index_name
-        expression /^(?<prefix>(?:container|router)-logs-e2e)-[^-]{10}-[^-]{26}_(?<suffix>.+)$/
+        expression /^(?<prefix>container-logs-e2e)-[^-]{10}-[^-]{26}_(?<suffix>.+)$/
         replace \\k<prefix>_\\k<suffix>
       </replace>
       <replace>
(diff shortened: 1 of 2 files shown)


smlx (commented):
Nice find!

You also might be able to solve this by just adding the other tag pattern to the existing filter block like this:
```
    <filter lagoon.*.container lagoon.*.router.openshift.**>
```

What do you think?

Schnitzel:
@smlx yea, would work as well, but I felt the way we do it here is a bit cleaner for anybody else to understand what happens to the router logs? also the second search/replace (for the apb) is not needed for the router logs, and as we have sooooo many router logs this might be a tiny bit faster?

charts/lagoon-logging/templates/logs-dispatcher.fluent-conf.configmap.yaml
```diff
@@ -270,7 +270,6 @@ data:
       skip_container_metadata true
       skip_master_url         true
     </filter>
-{{- end }}
```
smlx, line 273: Hey @Schnitzel this might be a mistake because it captures the `index_name` filter block inside the `.Values.openshiftHaproxyLogsCollector.enabled` block. Instead I think that the filter block adding `index_name` should go _after_ the `openshiftHaproxyLogsCollector` block?

smlx (changes requested):

charts/lagoon-logging/templates/logs-dispatcher.fluent-conf.configmap.yaml
smlx: Oh I think I get it.. `router-logs-e2e` aren't a thing on Kubernetes?

PR #172, opened 2020-12-03
add postgres-dbaas to makefile for cluster testing

tobybellwood:
This doesn't require a version bump, as it is only a makefile change that affects the services brought up with a lagoon-remote install.

### DIFF
--- a/charts/lagoon-remote/Chart.yaml
+++ b/charts/lagoon-remote/Chart.yaml
@@ -18,7 +18,7 @@ type: application
 # This is the chart version. This version number should be incremented each
 # time you make changes to the chart and its templates, including the app
 # version.
-version: 0.9.1
+version: 0.10.1
 
 # This is the version number of the application being deployed. This version
 # number should be incremented each time you make changes to the application.
--- a/charts/lagoon-remote/Chart.yaml
+++ b/charts/lagoon-remote/Chart.yaml
@@ -18,7 +18,7 @@ type: application
 # This is the chart version. This version number should be incremented each
 # time you make changes to the chart and its templates, including the app
 # version.
-version: 0.9.0
+version: 0.9.1
 
 # This is the version number of the application being deployed. This version
 # number should be incremented each time you make changes to the application.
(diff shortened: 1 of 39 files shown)


tobybellwood:
ah - this will fail until I merge https://github.com/amazeeio/lagoon/pull/2311 into main to fix up the tests image (I think)

tobybellwood:
@smlx - I've removed the tests from this PR, as they won't work until https://github.com/amazeeio/lagoon/pull/2311 is merged (which won't work until this PR is merged!)  Once this PR is merged here, I'll get the lagoon one passing, and then submit a new PR here for the tests.  phew...

charts/lagoon-remote/Chart.yaml
```diff
@@ -18,7 +18,7 @@ type: application
 # This is the chart version. This version number should be incremented each
 # time you make changes to the chart and its templates, including the app
 # version.
-version: 0.10.0
+version: 0.10.1
```
smlx, line 21: This isn't required since there's no longer any change to the chart.

smlx (approved):
LGTM. Since this is now only touching the `Makefile` there isn't even a chart release.

I'd say just remove the version bump on `lagoon-remote` and merge!

smlx:
@tobybellwood Here's how I've been doing these inter-dependent merges on Lagoon and the Charts.

1. Make the change required in the charts repo and push a branch.
2. Push a PR to Lagoon which also temporarily changes the `CHARTS_TREEISH` to your branch name from 1. Get this green, and merge it.
3. Create a PR in Charts repo, get it green and merge it.
4. Create PR back in Lagoon pointing at the tag from 3.

It's a bit complex but keeps each repo "locked" on a known-working version of the other.

charts/lagoon-remote/Chart.yaml
tobybellwood: It was needed when I had a test in there, but not anymore

tobybellwood:
I'll test it using CHARTS_TREEISH anyway first :-)

smlx (approved):

Issue #151, opened 2020-11-10
Remove update-push-images.sh from lagoon-remote chart


xantrix:
In k8s we do not need anymore to push to the Openshift registry
https://github.com/amazeeio/lagoon/blob/main/images/docker-host/update-push-images.sh
so we can remove it from Deployment chart https://github.com/uselagoon/lagoon-charts/blob/main/charts/lagoon-remote/templates/docker-host.deployment.yaml#L46

This will solve the following issue in the k8s clusters.

```bash
go-crond: failed cronjob: cmd:/lagoon/cronjob.sh "/update-push-images.sh" out:2020-11-10T15:15:00Z CRONJOB: /update-push-images.sh
++ cat /var/run/secrets/kubernetes.io/serviceaccount/namespace
+ NAMESPACE=lagoon
+ docker -H localhost info
+ docker login -u=serviceaccount --password-stdin registry.lagoon.svc:5000
Error response from daemon: Get http://registry.lagoon.svc:5000/v2/: dial tcp: lookup registry.lagoon.svc on 172.20.0.10:53: no such host
 err:exit status 1 time:166.646926m
```

smlx:
thanks for finding this!

I'm not 100% sure what the purpose of this script is. If you can confirm that it's not needed on k8s did you want to create a PR?

xantrix:
hi @smlx 
I'm going to have a deeper look into it and of course happy to make a PR :)

PR #173, opened 2020-12-03
add update-images.sh for docker-host

xantrix:
Linked PR in Lagoon
https://github.com/amazeeio/lagoon/pull/2368

Closes: #151

### DIFF
--- a/charts/lagoon-remote/values.yaml
+++ b/charts/lagoon-remote/values.yaml
@@ -19,7 +19,7 @@ dockerHost:
 
   registry: registry.lagoon.svc:5000
 
-  repositoryToUpdate: lagoon
+  repositoryToUpdate: amazeeio|lagoon
 
   replicaCount: 1
 
--- a/charts/lagoon-remote/Chart.yaml
+++ b/charts/lagoon-remote/Chart.yaml
@@ -18,7 +18,7 @@ type: application
 # This is the chart version. This version number should be incremented each
 # time you make changes to the chart and its templates, including the app
 # version.
-version: 0.17.0
+version: 0.17.1
 
 # This is the version number of the application being deployed. This version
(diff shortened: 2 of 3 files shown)


charts/lagoon-remote/values.yaml
```diff
@@ -29,7 +29,7 @@ dockerHost:
 
   registry: registry.lagoon.svc:5000
 
-  repositoryToUpdate: lagoon
```
smlx, line 32: I think this change might not work.

Looking at the script in https://github.com/amazeeio/lagoon/pull/2368 it greps for `${REPOSITORY_TO_UPDATE}/`. Our base image repositories are [`testlagoon`](https://hub.docker.com/u/testlagoon) and [`uselagoon`](https://hub.docker.com/u/uselagoon), so having `amazeeio` here will mean that the base images don't get updated. Could you confirm that understanding of where the base images are pushed @tobybellwood?

tobybellwood: It's an -E grep, so it probably should cover all our repos? 
`amazeeio|uselagoon` for production clusters, and add `testlagoon` if there's going to be test workloads on it?

`amazeeio|lagoon` would also cover it

xantrix: thanks for clarifying @tobybellwood 
in our production cluster, I saw `amazeeio` as base image repo.
If we agree I'd proceed using `amazeeio|lagoon`

smlx: sounds good :+1:

smlx:
hi @xantrix sorry this has languished a bit.. could you rebase onto main and we'll get this in ASAP.

xantrix:
no worries @smlx :) rebased

smlx (approved):
LGTM

PR #174, opened 2020-12-03
Consolidation fixes

smlx:
While reviewing #156 and debugging some other error messages in the `logs-concentrator` I realised that the consolidation syntax was wrong.

Going back to my original testing for this I discovered that this is actually an issue with the upstream fluentd `record_modifier` plugin documentation and test suite. There's a PR for upstream here that explains the issue: https://github.com/repeatedly/fluent-plugin-record-modifier/pull/56

So I incorporated the changes for #156 and also fixed the syntax. I hope you don't mind that I took some config snippets from your PR @Schnitzel!

Supersedes / Closes: #156

### DIFF
--- a/charts/lagoon-logging/Chart.yaml
+++ b/charts/lagoon-logging/Chart.yaml
@@ -18,7 +18,7 @@ type: application
 # time you make changes to the chart and its templates, including the app
 # version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.17.0
+version: 0.17.1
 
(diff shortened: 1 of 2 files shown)


smlx:
@Schnitzel just to confirm - those `e2e` namespaces are only on openshift?

Schnitzel:
@smlx 
yes correct, they are only existing in OpenShift as they are part of the OpenShift monitoring infrastructure we run

Schnitzel (approved):

PR #186, opened 2020-12-20
use uselagoon for lagoon-core

Schnitzel:


charts/lagoon-core/values.yaml
```diff
@@ -595,7 +595,7 @@ drushAlias:
 logsDBCurator:
   enabled: true
   image:
-    repository: amazeeiolagoon/logs-db-curator
+    repository: uselagoon/logs-db-curator
     pullPolicy: Always
     # Overrides the image tag whose default is the chart appVersion.
     tag: ""
```
tobybellwood, line 601: Until https://github.com/amazeeio/lagoon/pull/2409 is released - we have to use the testlagoon version of this image
```suggestion
    repository: testlagoon/logs-db-curator
    pullPolicy: Always
    # Overrides the image tag whose default is the chart appVersion.
    tag: "pr-2409"
```

charts/lagoon-core/values.yaml
```diff
@@ -639,7 +639,7 @@ sshPortal:
   enabled: false
   replicaCount: 2
   image:
-    repository: amazeeiolagoon/ssh-portal
+    repository: uselagoon/ssh-portal
```
tobybellwood, line 642: pr-branches are only in testlagoon
```suggestion
    repository: testlagoon/ssh-portal
```

tobybellwood:
This is the first step towards the larger PR in #183 .

Note that the `main` imageTag doesn't exist in the `uselagoon` repo - so any cluster that builds off main will also need to provide `testlagoon` based repository overrides in their values files (or we push main branch into the `uselagoon` namespace too?)

Schnitzel:
@tobybellwood 
oh wow, sorry didn't see #183 when I pushed that one.

Let's close this one and work on #183

PR #195, opened 2020-12-29
lagoon-remote docker-host openshift support

Schnitzel:
Explain the **details** for making this change. What existing problem does the pull request solve?

Put `Closes: #XXXX` in your comment to auto-close the issue that your PR fixes (if such).

### DIFF
--- a/charts/lagoon-remote/templates/docker-host.clusterrole.yaml
+++ b/charts/lagoon-remote/templates/docker-host.clusterrole.yaml
@@ -1,0 +1,17 @@
+{{- if and .Values.dockerHost.serviceAccount.create .Values.global.openshift -}}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "lagoon-remote.dockerHost.fullname" . }}
+  labels:
+    {{- include "lagoon-remote.dockerHost.labels" . | nindent 4 }}
+rules:
+- apiGroups:
+  - security.openshift.io
+  resourceNames:
+  - privileged
+  resources:
+  - securitycontextconstraints
+  verbs:
+  - use
+{{- end }}
--- a/charts/lagoon-remote/values.yaml
+++ b/charts/lagoon-remote/values.yaml
@@ -1,5 +1,10 @@
 # Default values for lagoon-remote.
 
+# global values which can affect subcharts
+global:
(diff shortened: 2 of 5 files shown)


charts/lagoon-remote/values.yaml
```diff
@@ -34,6 +34,8 @@ dockerHost:
     # If not set and create is true, a name is generated using the fullname
     # template
     name:
+    # set to true if you deploy the dockerhost to an openshift
+    isOpenshift: false
```
smlx, line 38: we should probably make this a global variable since other services/subcharts (e.g. dioscuri) have a similar flag.

charts/lagoon-remote/values.yaml
```diff
@@ -110,7 +112,7 @@ dioscuri:
 
   resources: {}
 
-  # this sidecar runs in the same pod as dioscuri
+  # this sidecar runs in the same pod as dioscurid
```
smlx, line 115: typo?

charts/lagoon-remote/templates/docker-host.clusterrole.yaml
```diff
@@ -0,0 +1,17 @@
...
+rules:
+- apiGroups:
+  - security.openshift.io
+  resourceNames:
+  - privileged
+  resources:
+  - securitycontextconstraints
+  verbs:
+  - use
+{{- end }}
```
smlx, line 17: missing a newline here and in the rolebinding template

smlx (changes requested):

charts/lagoon-remote/values.yaml
Schnitzel: yea, thought about this too, any idea how to do this? as here we need this as boolean, but the other charts need it as an additional entry in a yaml list, plus we cannot use golang templating in the values.yaml file :)

smlx: since we control both the parent and subchart we can use a [global value](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#global-chart-values) like `.Values.global.openshift` and then do something like `coalesce .Values.openshift .Values.global.openshift` in the templates. What do you think?

Schnitzel: sounds good to me!

smlx (approved):

PR #198, opened 2021-01-05
Collect logs from system namespaces

Schnitzel:
We had a couple of cases where we really needed logs from system namespaces.
I'm not sure though if this is really the right way to do this. Basically we're sending many thousand of log entries to the elasticsearch which is also used by our customers. 
If I would have time I probably would create a separate logging infrastructure for system logs from our clusters, but as we don't really have time I think the easiest right now is to just send them in the same place.
Any opinions @smlx ?

### DIFF
--- a/charts/lagoon-logging/Chart.yaml
+++ b/charts/lagoon-logging/Chart.yaml
@@ -18,7 +18,7 @@ type: application
 # time you make changes to the chart and its templates, including the app
 # version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.19.0
+version: 0.20.0
 
 # This is the version number of the application being deployed. This version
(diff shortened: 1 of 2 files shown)


smlx:
Here are some problems I see with doing this:
* many system services are very verbose, this would add significant load on elasticsearch
* fluentd currently looks for `lagoon.sh/project` labels and otherwise puts all logs into a catch-all index, I'm not sure how well parsing will work etc.

I agree that it would be much better to split these out to a different log sink. However that doesn't need to be elasticsearch and could be something much simpler since we don't need the multi-tenancy. I've got some ideas for this :smile: 

Another thought: do we really need _all_ these namespaces? one or two might be ok as a quick fix for your use-case?

PR #210, opened 2021-02-03
Lagoon core user consistency

smlx:
Use consistent user/group IDs for workloads in lagoon-core too.

This isn't enforced, just better than using some other random IDs.

~Draft until the associated PR is accepted into amazeeio/lagoon.~

PR #217, opened 2021-02-10
add needs-testing label logic to github actions

tobybellwood:
This PR adds a logic step to the k8s cluster steps that will perform them even if there are no chart changes.

This may be needed to test logic in the makefiles etc.  To use it, allow the skipped run of the GitHub action to complete, then add the `needs-testing` label, and trigger a new run for the PR.

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -84,7 +84,7 @@ jobs:
     - name: Create kind cluster
       uses: helm/kind-action@v1.1.0
       if: |
-        (steps.list-changed.outputs.changed == 'true') || 
+        (steps.list-changed.outputs.changed == 'true') ||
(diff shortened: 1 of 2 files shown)


smlx:
Hmm. I followed the instructions and it didn't fire. Any ideas @tobybellwood?

tobybellwood:
> Hmm. I followed the instructions and it didn't fire. Any ideas @tobybellwood?

I don't think the label will trigger a new action itself - but can't recall if you can do it via the UI or if you need to push a blank commit

smlx:
I did re-run the jobs on this PR. Would have thought that might do it? :thinking:

smlx:
ah, the new commit seems to have done the job!

smlx:
rebased to get the fixes in https://github.com/uselagoon/lagoon-charts/pull/210 so CI passes

PR #222, opened 2021-02-15
chore: remove unused variable from Makefile

smlx:
This doesn't appear to be used in any charts.


tobybellwood (approved):
Agree - that was only in there until we'd changed all the variables to align again.

PR #226, opened 2021-02-19
Support rootless build pods

smlx:
This PR relates to https://github.com/amazeeio/lagoon/pull/2523.

It depends on:
* [x] approval of the change in https://github.com/amazeeio/lagoon-kbd/pull/38
* [x] a lagoon-kbd release with that change


smlx:
CI works with this Lagoon change: https://github.com/amazeeio/lagoon/pull/2523

tobybellwood:
Technically this branch isn't needed, right - the controller already supports rootless build pods, and there is a values.yaml switch already to activate them. We would only commit these changes here if we wanted to test rootless builds?

smlx:
Yes correct.

PR #233, opened 2021-03-05
fix wrong s3FilesSecretAccessKey name

Schnitzel:
was called `s3SecretAccessKey` in the values.yaml, but the template actually searches for `s3FilesSecretAccessKey`


smlx (approved):
Thanks!

PR #237, opened 2021-03-17
Create external ingress for webhook and configure URL for Harbor scans

tobybellwood:
This PR (alongside https://github.com/amazeeio/lagoon/pull/2580) re-enables the sending of webhooks from Harbor to Lagoon.

It does this by creating an ingress for the webhook service (along the lines of the registry one) and then setting the webhookURL in the relevant services.

This PR also renames the kubernetes namespace for harbor (back to harbor), as there are Lagoon internals that are set to it as a default.  It doesn't change the external registry.* endpoint.

This PR can be merged irregardless of the Lagoon one - it's backwards compatible too.


smlx (approved):
LGTM

smlx:
@tobybellwood is this PR still required?

tobybellwood:
nope!

PR #238, opened 2021-03-19
Move default args to template for builddeploy controller

shreddedbacon:
When needing to add `extraArgs` to the builddeploy controller, defining new values in an external template would override the defaults provided in the template, so they need to be defined again wherever the `extraArgs` are defined.

This moves them into the template and frees up `extraArgs` for actual extra arguments.

### DIFF
--- a/charts/lagoon-build-deploy/templates/deployment.yaml
+++ b/charts/lagoon-build-deploy/templates/deployment.yaml
@@ -47,6 +47,8 @@ spec:
         - /manager
         {{- if (or .Values.rootlessBuildPods .Values.extraArgs) }}
         args:
+        - "--metrics-addr=127.0.0.1:8080"
+        - "--enable-leader-election=true"
         {{- if .Values.rootlessBuildPods }}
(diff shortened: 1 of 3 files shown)


charts/lagoon-build-deploy/templates/deployment.yaml
```diff
@@ -47,6 +47,8 @@ spec:
         - /manager
         {{- if (or .Values.rootlessBuildPods .Values.extraArgs) }}
```
smlx, line 48: I think this logic won't work anymore with `extraArgs` being empty. In that case there will be no arguments at all to `/manager`.

I think it will work if you just remove this outer `if` statement.

smlx (changes requested):

charts/lagoon-build-deploy/templates/deployment.yaml
shreddedbacon: 🤦 ah yea. Will fix

smlx (approved):

PR #244, opened 2021-03-25
allow to control fluentd replicas from values.yaml

tobybellwood:
This PR allows helm templates to control the replicas for the fluentd pods inside the logging operator.  It keeps the default at 3, but uses 1 for CI (and we would also use 1 for local development)

### DIFF
--- a/charts/lagoon-logging/ci/linter-values.yaml
+++ b/charts/lagoon-logging/ci/linter-values.yaml
@@ -44,6 +44,7 @@ tls:
     FejDrD4asORcLF9xN2tsM5fGoBuHy9KpUw==
     -----END EC PRIVATE KEY-----
 
+fluentdreplicaCount: 1
 consolidateServiceIndices: true
 
 logging-operator:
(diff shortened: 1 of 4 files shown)


smlx (approved):
LGTM

charts/lagoon-logging/ci/linter-values.yaml
```diff
@@ -44,6 +44,7 @@ tls:
     FejDrD4asORcLF9xN2tsM5fGoBuHy9KpUw==
     -----END EC PRIVATE KEY-----
 
+fluentdreplicaCount: 1
```
smlx, line 47: I just realised the capitalisation is not correct here so you'll still get the default `3`.

smlx (changes requested):

charts/lagoon-logging/ci/linter-values.yaml
tobybellwood, line 47:
```suggestion
fluentdReplicaCount: 1
```

PR #245, opened 2021-03-26
Update builddeploy chart to use latest image

shreddedbacon:
Update the builddeploy chart to use the latest image of [v0.1.8 (changes)](https://github.com/amazeeio/lagoon-kbd/releases/tag/v0.1.8)

Also updated the CRD to support https://github.com/amazeeio/lagoon-kbd/pull/41 as the second part of the fix for https://github.com/amazeeio/lagoon/issues/2568

### DIFF
--- a/charts/lagoon-build-deploy/Chart.yaml
+++ b/charts/lagoon-build-deploy/Chart.yaml
@@ -15,6 +15,6 @@ maintainers:
 
 type: application
 
-version: 0.5.0
+version: 0.6.0
 
-appVersion: v0.1.7
+appVersion: v0.1.8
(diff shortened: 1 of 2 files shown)


smlx (approved):
This looks fine, but it will need some manual intervention when upgrading because the CRD will need to be manually applied. `helm` won't automatically re-apply the version to make the new fields usable - it will need `helm apply -f ...`.

We'll need to note that somewhere because I guess the controller v0.1.8 will try to populate those fields in custom resources?

shreddedbacon:
It shouldn't break the controller if the CRD is not updated, just the thing it is supposed to fix won't be fixed.

I don't know what the current process is with applying updates to clusters when updated CRDs are involved. If we haven't already got a plan for this, now might be the time to get all required parties involved to decide the best way forward?

PR #246, opened 2021-04-03
Feature/adds logs2webhooks

bomoko:


bomoko:
Going to be pushing this to the repo directly and reopening

PR #249, opened 2021-04-12
Update builddeploy controller to v0.1.9

shreddedbacon:
This is to replace #245 so the same issues with the CRD update will apply here. (https://github.com/uselagoon/lagoon-charts/pull/245#pullrequestreview-625133433)


smlx (approved):
LGTM

PR #252, opened 2021-05-04
Docker host storageclass

### LINKED ISSUE #251: Make docker-host PVC storageclass configurable
Currently Azure caps the standard storageclass at 500 IOPS. The docker-host needs more than that, so it should use one of the premium storage classes.

This should be either the default on Azure somehow, or at least configurable.

smlx:
This PR makes the docker-host PVC storage class configurable through the helm chart values. By default we keep the current behaviour of not defining a storage class so as to fall back to cluster default.

Closes #251

### DIFF
--- a/charts/lagoon-remote/Chart.yaml
+++ b/charts/lagoon-remote/Chart.yaml
@@ -18,7 +18,7 @@ type: application
 # This is the chart version. This version number should be incremented each
 # time you make changes to the chart and its templates, including the app
 # version.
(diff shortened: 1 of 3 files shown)


smlx:
this passed here: https://lagoon-ci.amazeeio.cloud/blue/organizations/jenkins/lagoon/detail/dockerhost-sc/1/pipeline/50

will fix CI here https://github.com/uselagoon/lagoon-charts/issues/253

PR #254, opened 2021-05-07
lagoon v2.0.0-alpha.9 release update appVersion

tobybellwood:


smlx:
this looks good. did you figure out why `drupal-php73` was failing?

tobybellwood:
I think it just takes too long - for some reason the php73 image takes longer than PHP 74, and this test does two full runs

smlx (approved):

PR #258, opened 2021-05-13
fix StatefulSet typo in logs-concentrator

tobybellwood:
version increment has been set to a minor behind #254 which should be merged first.


smlx (approved):
LGTM

PR #259, opened 2021-05-14
add JWTSECRET to test-suite

tobybellwood:
This PR adds the JWTSECRET to the lagoon-test-test-suite.

This will enable ansible playbooks etc to perform GraphQL mutations in tests to add and delete projects, groups, notifications, and other permissions not available to normal users.

### DIFF
--- a/charts/lagoon-test/Chart.yaml
+++ b/charts/lagoon-test/Chart.yaml
@@ -10,6 +10,6 @@ maintainers:
 
 type: application
 
-version: 0.14.0
+version: 0.14.1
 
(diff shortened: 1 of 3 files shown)


smlx (commented):
Should this variable maybe just go into the `secret.yaml` where we already have the other variables?

tobybellwood:
I considered that but figured I only wanted it in the one place.  Plus it seemed harder to work out how to put a secret secret into a secret :smile:

smlx:
Ah sorry, I see what you mean. We already do it this way in the data-watcher-pusher. :+1:

Issue #262, opened 2021-05-28
Error when settings GITLAB_API_TOKEN 


dwoods:
Setting `gitlabAPIToken' is causing the following error:

```
executing "lagoon-core/templates/webhooks2tasks.deployment.yaml" at <include "lagoon-core.api.fullname" .>: 
error callinginclude: template: lagoon-core/templates/_helpers.tpl:55:4: executing "lagoon-core.api.fullname" 
at <include "lagoon-core.fullname" .>: error calling include: template: lagoon-core/templates/_helpers.tpl:15:22: 
executing "lagoon-core.fullname" at <.Chart.Name>: can't evaluate field Chart in type string
```

I believe this is due to the fact that the '.' in the {{- with ... }} block has a different value in that context.

Will submit a pull request for the fix that worked for me.

PR #263, opened 2021-05-28
Gitlab API token fix

dwoods:
This pull request fixes an issue where settings the gitlabAPIToken helm value causes an error rendering the chart yaml.  The {{ include }} for the secret name fails due to it being in a 'with' block.

Closes: #262

### DIFF
--- a/charts/lagoon-core/Chart.yaml
+++ b/charts/lagoon-core/Chart.yaml
@@ -5,9 +5,9 @@ description: >
 home: https://github.com/uselagoon/lagoon-charts
 icon: https://raw.githubusercontent.com/uselagoon/lagoon-charts/main/icon.png
 maintainers:
-- name: smlx
-  email: scott.leggett@amazee.io
-  url: https://amazee.io
+  - name: smlx
+    email: scott.leggett@amazee.io
+    url: https://amazee.io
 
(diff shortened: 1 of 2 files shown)


charts/lagoon-core/Chart.yaml
```diff
@@ -5,9 +5,9 @@ description: >
 home: https://github.com/uselagoon/lagoon-charts
 icon: https://raw.githubusercontent.com/uselagoon/lagoon-charts/main/icon.png
 maintainers:
-- name: smlx
-  email: scott.leggett@amazee.io
-  url: https://amazee.io
+  - name: smlx
```
smlx, line 8: I think this might be a mistake.. could we change it back?

smlx (changes requested):

charts/lagoon-core/Chart.yaml
smlx: otherwise looks good!

dwoods: Reverted changes, thanks!

smlx:
hey @dwoods happy to merge this if you can bump the chart version again

dwoods:
Bumped chart version

smlx:
Thanks @dwoods!  :tada:

Issue #271, opened 2021-06-23
Calico causes problems on Docker for Mac


smlx:
Calico causes networking instability in Docker for Mac.

We should disable Calico by default and only use it for the Network Policy related tests.

PR #272, opened 2021-06-23
Make Calico CNI plugin optional

smlx:
Make Calico opt-in and rely on the default kindnet otherwise.

This should improve the local testing experience, particularly on Mac.

Closes: #271


tobybellwood (approved):
this worked locally for me in lagoon-main via CHARTS_TREEISH

PR #278, opened 2021-06-28
Phonify more targets in the Makefile

smlx:
This is a minor nitpick, but just makes all targets consistent.


tobybellwood (approved):
Sorry, my bad!

PR #279, opened 2021-06-28
Remove install-tests target from Makefile

smlx:
The install-tests target was installing the lagoon-test chart manually, which breaks assumptions in the helm chart-testing tooling and also breaks `kind/retest` in the `uselagoon/lagoon` Makefile.

So remove that target and instead update the fill-test-ci-values target to do the thing that install-tests was intended to do.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,8 +26,13 @@ LAGOON_FEATURE_FLAG_DEFAULT_ISOLATION_NETWORK_POLICY =
 # is useful for testing network policies.
 USE_CALICO_CNI =
 # Set to `true` to assume that `make install-registry` has been run manually.
-# This avoids running install-registry twice in uselagoon/lagoon CI.
+# This avoids running install-registry twice in uselagoon/lagoon CI when
+# invoking fill-test-ci-values.
 SKIP_INSTALL_REGISTRY =
+# Set to `true` to assume that all dependencies have already been installed.
+# This allows updating the fill-test-ci-values template only, without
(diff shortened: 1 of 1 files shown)


smlx:
I've got an associated PR in the works for the Lagoon Makefile and docs.

tobybellwood:
Maybe the `install-tests` routine name should be preserved to bundle `fill-test-ci-values` and it's pre-reqs (install-ingress install-lagoon-core install-lagoon-remote install-nfs-server-provisioner) rather than have the fill-test-ci-values performing dual tasks based on a var - IMO it makes sense to just use it to fill values?  We'd obvs have to change a couple other actions tho?

smlx:
The thing is that in order to fill the template values those charts _do_ actually have to be installed, so IMO it makes sense to declare the dependencies. It's just that the special case of `kind/retest` in the Lagoon Makefile is reusing the existing cluster.

This complexity is all hidden from the user of the Makefile so I though it made sense to keep the logic for templating the yaml in one place?

I'll send some PRs to Lagoon shortly with some associated documentation and other tweaks to the Makefile that might help show where I'm going with this. Happy for you to rework this if you like of course :slightly_smiling_face:

tobybellwood:
I'm still happy, just wanted to check - I just always thought it strange that a fill-test-ci-values command was used to deploy an entire Lagoon!

tobybellwood (approved):
I'm happy - it'll be better for getting the retest logic smarter!

PR #285, opened 2021-06-30
allow to inject additional env variables into services

Schnitzel:
many of our lagoon services have env variables to control them, while it doesn't make sense to expose each of them via individual helm variables, we can provide a possibility to inject them this way

### DIFF
--- a/charts/lagoon-core/Chart.yaml
+++ b/charts/lagoon-core/Chart.yaml
@@ -17,7 +17,7 @@ type: application
 # time you make changes to the chart and its templates, including the app
 # version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.44.0
+version: 0.45.0
 
 # This is the version number of the application being deployed. This version
 # number should be incremented each time you make changes to the application.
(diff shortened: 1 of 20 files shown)


tobybellwood (approved):
Nice change - this will allow for much more customisation without changing the charts themselves!

smlx:
I am not sure that this is a good idea because:
* these interfaces are undocumented and not discoverable
* variables cannot be shared between services and must be defined individually for each service
* it encourages hard-coding of variables which might be better namespaced. e.g. `NEW_RELIC_APP_NAME: lagoon-api-<cluster_name>` might be better than `NEW_RELIC_APP_NAME: lagoon-api`?
* we can probably auto-generate some of these and/or set sensible defaults (e.g. `API_DB_CONN_LIMIT`)
* it makes it more difficult to catch errors using a [values schema](https://helm.sh/docs/topics/charts/#schema-files)

@tobybellwood / @Schnitzel thoughts on adding the few extra values fields for these variables instead?

Schnitzel:
@smlx 
I'm ok with adding them as actual helm variables over time, but I really needed a way right now to get our production system stable again, as we have injected variables via manually editing deploymentconfigs and they are overwritten on every helm run, so we lost some variables that we've added before, causing outages and unhappyness of customers.

I think we would need a discovery to find all variables that each service supports and then add them all to the helm values. If somebody want's to do that, feel free to work on this. I'm currently more focused on having a stable platform :D

PR #290, opened 2021-07-02
inject API_HOST as url and not just host

Schnitzel:


tobybellwood (approved):
Sorry I didn't get this into the release!

PR #300, opened 2021-07-12
Adding k8up annotations to api-db and keycloak-db pods in lagoon-core chart

cdchris12:
This PR adds support for the annotations k8up needs to have to backup the `api-db` and `keycloak-db` pods' PVCs in the lagoon-core chart

### DIFF
--- a/charts/lagoon-core/templates/api-db.statefulset.yaml
+++ b/charts/lagoon-core/templates/api-db.statefulset.yaml
@@ -13,10 +13,8 @@ spec:
     metadata:
       annotations:
         checksum/api-db.secret: {{ include (print $.Template.BasePath "/api-db.secret.yaml") . | sha256sum }}
-        {{- if .Values.k8up }}
         k8up.syn.tools/backupcommand: /bin/sh -c "mysqldump --max-allowed-packet=500M --events --routines --quick --add-locks --no-autocommit --single-transaction --all-databases"
         k8up.syn.tools/file-extension: .api-db.sql
-        {{- end }}
     {{- with .Values.apiDB.podAnnotations }}
         {{- toYaml . | nindent 8 }}
     {{- end }}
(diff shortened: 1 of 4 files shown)


charts/lagoon-core/templates/api-db.statefulset.yaml
```diff
@@ -13,6 +13,10 @@ spec:
     metadata:
       annotations:
         checksum/api-db.secret: {{ include (print $.Template.BasePath "/api-db.secret.yaml") . | sha256sum }}
+        {{- if .Values.k8up }}
+        appuio.ch/backupcommand: /bin/sh -c "mysqldump --max-allowed-packet=500M --events --routines --quick --add-locks --no-autocommit --single-transaction --all-databases"
```
Schnitzel, line 17: let's use the default annotations `k8up.syn.tools/backupcommand` and `k8up.syn.tools/file-extension` of k8up, see https://k8up.io/k8up/1.1.0/references/config-reference.html

charts/lagoon-core/values.yaml
```diff
@@ -50,6 +50,8 @@
 
 # These values are the chart defaults, but can be overridden.
 
+k8up: false
```
Schnitzel, line 53: as these are just annotations and don't hurt, I would make them enabled by default.

charts/lagoon-core/templates/keycloak-db.statefulset.yaml
```diff
@@ -13,6 +13,10 @@ spec:
     metadata:
       annotations:
         checksum/keycloak.secret: {{ include (print $.Template.BasePath "/keycloak.secret.yaml") . | sha256sum }}
+        {{- if .Values.k8up }}
+        appuio.ch/backupcommand: /bin/sh -c "mysqldump --max-allowed-packet=500M --events --routines --quick --add-locks --no-autocommit --single-transaction --all-databases"
+        backup.appuio.ch/file-extension: .api-db.sql
```
Schnitzel, line 18: this is keycloak, so let's call it `.keycloak-db.sql`

Schnitzel (changes requested):

charts/lagoon-core/templates/api-db.statefulset.yaml
cdchris12: Good catch; I just copypasta'd those a bit too quickly

smlx (approved):
LGTM, thanks!

Schnitzel (approved):

PR #307, opened 2021-07-22
Elasticsearch  URL & Host Mess

Schnitzel:
Our values for elasticsearch is a mess, we have multiple ways to define the same thing.
This change now just uses `elasticsearchURL` like we already have `kibanaURL` and other `xxxURL`
Turns out that all services in lagoon core can handle full URLs with Scheme and Ports (like they should), so it's probably the best to do it this way. 
As soon as https://github.com/uselagoon/lagoon/pull/2786 is merged we can also remove `ELASTICSEARCH_HOST` but until then I would leave it in for BC


tobybellwood:
Merged into #318 with the release that removes ELASTICSEARCH_HOST from lagoon-core.

Note that logs-concentrator still uses ELASTICSEARCH_HOST in this chart (and presumably other ones upstream)

PR #308, opened 2021-07-22
remove logs-db-curator, it's not used anymore since we switched to opendistro

Schnitzel:
also see sister PR for lagoon: https://github.com/uselagoon/lagoon/pull/2787


smlx (approved):
LGTM

PR #309, opened 2021-07-23
Adding k8up schedule support for lagoon-core

cdchris12:
This PR adds support for backing up all objects in the `lagoon-core` chart with k8up via the addition of a `Schedule` object and a `baas-repo-password`


smlx:
Hey @cdchris12 could you bump the version again to get CI running?

cdchris12:
> Hey @cdchris12 could you bump the version again to get CI running?

Sure; pushed a new commit up just now. :)

charts/lagoon-core/templates/k8up.schedule.yaml
```diff
@@ -0,0 +1,24 @@
+{{ if .Capabilities.APIVersions.Has "backup.appuio.ch/v1alpha1/Schedule" }}
+apiVersion: backup.appuio.ch/v1alpha1
+kind: Schedule
+metadata:
+  name: 'lagoon-core-schedule'
```
smlx, line 5: could we change this to not be hard-coded? that way it is consistent with the other templates.
```suggestion
  name: {{ include "lagoon-core.fullname" . }}
```

charts/lagoon-core/templates/k8up.secret.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: 'baas-repo-pw'
```
smlx, line 4: similar to above I think we should have
```suggestion
  name: "{{ include "lagoon-core.fullname" . }}-baas-repo-pw"
```

charts/lagoon-core/templates/k8up.schedule.yaml
```diff
@@ -0,0 +1,24 @@
...
+spec:
+  backend:
+    repoPasswordSecretRef:
+      key: repo-pw
+      name: baas-repo-pw
```
smlx, line 10:
```suggestion
      name: "{{ include "lagoon-core.fullname" . }}-baas-repo-pw"
```

charts/lagoon-core/values.yaml
```diff
@@ -3,6 +3,7 @@
 # elasticsearchHost:
 # harborAdminPassword:
 # harborURL:
+# k8upBackupBucketName:
```
smlx, line 6: Could we make this value optional so that the bucket name can be autogenerated by default? e.g. generate a bucket name from something like `{{ include "lagoon-core.fullname" . }}-{{ randAlphaNum 4 }}` if the value isn't set, but allow it to be overridden if this values is set?

smlx (changes requested):

charts/lagoon-core/templates/k8up.schedule.yaml
```diff
@@ -0,0 +1,33 @@
...
+    s3:
+      {{ if .Values.k8upBackupBucketName }}
+      bucket: '{{ .Values.k8upBackupBucketName }}'
+      {{ else if (lookup "backup.appuio.ch/v1alpha1" "Schedule" "" (include "lagoon-core.fullname" .) )}}
+      bucket: {{ (lookup "backup.appuio.ch/v1alpha1" "Schedule" "" (include "lagoon-core.fullname" .) ).spec.backend.s3.bucket }}
+      {{ else }}
+      bucket: {{ include "lagoon-core.fullname" . }}-{{ randAlphaNum 8 }}
+      {{ end }}
+      {{ if .Values.k8upS3Endpoint }}
+      endpoint: {{ include .Values.k8upS3Endpoint}}
```
smlx, line 20: very minor: I don't think the `include` is required.

smlx (approved):
Nice one!

PR #312, opened 2021-08-05
conditionally checking the secret, not the release.IsUpgrade for k8up…

timclifford:
… secret


smlx (approved):
yup, this is much nicer :)

smlx:
Tested locally, this fixes the upgrade. :tada:

PR #326, opened 2021-09-14
lowercasing the generated lagoon-core k8up bucket name

cdchris12:
This PR adds support for casting the generated bucket name for lagoon-core's k8up schedule into lowercase characters and numbers only. This is required because uppercase characters are not allowed to be used as an S3 bucket name.


smlx (approved):
LGTM

PR #329, opened 2021-09-23
Adding "baas-" prefix to lagoon-core backup bucket

cdchris12:
This PR adds the `baas-` prefix to the bucket name generated for the lagoon-core chart. This is required for the permissions set up for the `k8up` pod's IAM keys to be allowed to access this bucket.


smlx (approved):
LGTM

PR #330, opened 2021-09-30
Add secret annotation to auth-server to make sure it is restarted

shreddedbacon:
All other services that use JWT_SECRET have an annotation to check if the secret has changed to force an update, except auth-server. This service is pretty important too.


smlx (approved):
LGTM

PR #356, opened 2021-11-25
docker-host: disable selinux relabeling

Schnitzel:
Ensures selinux relabeling is disabled for the docker-host, this would case the container never to start as there can be so many files in the persistent storage.

This only applies for k8s distributions that have selinux enabled (like OpenShift 4). Other k8s distributions should ignore this.

Example in `cri-o` where this is implemented:
https://github.com/cri-o/cri-o/blob/7048f2440217bbb2d6b5b7cfdbad052a4b8527f1/server/container_create_linux.go#L284-L288


smlx:
CI passes so I guess it doesn't break a non-selinux node :shipit:

PR #358, opened 2021-12-07
Fix formatting of the install steps in lagoon-logging README

rocketeerbkw:
The markdown was written in a way that splits all the ordered lists into separate `<ol>` elements, this isn't great for accessibility. I was also initially confused about the multiple "Openshift Only" sections.

This rewrite renders everything into a single `<ol>` which makes it clearer that the "Openshift Only" instructions are for the installation step they're listed under. I also started the list from 1 instead of 0 since this is a README, not code

### DIFF
--- a/charts/lagoon-logging/Chart.yaml
+++ b/charts/lagoon-logging/Chart.yaml
@@ -19,7 +19,7 @@ type: application
 # time you make changes to the chart and its templates, including the app
 # version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.59.0
+version: 0.59.1
 
 # This is the version number of the application being deployed. This version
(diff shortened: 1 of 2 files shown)


rocketeerbkw:
Side-by-side difference (incomplete, but illustrative)
lagoon-charts_README_md_at_d6cd83d71e842ecea7d60396cc35316dccbf8efa_·_uselagoon_lagoon-charts_and_lagoon-charts_charts_lagoon-logging_at_main_·_uselagoon_lagoon-charts_and_Document_additional_requirements_for_collecting_logs_of_the_NGINX_in

smlx:
This is great, thank you!

Could you just bump the chart version by one patch level in `Chart.yaml`?

PR #361, opened 2021-12-21
Rename lagoon-kbd to remote-controller in README

tobybellwood:


smlx:
@tobybellwood this needs a rebase I think?

PR #364, opened 2022-01-20
insights handler deployment

timclifford:
corresponding PR uselagoon/lagoon#2997


bomoko:
@timclifford - this LGTM, but I'd like someone with more Helm-foo to look at it too.

charts/lagoon-core/values.yaml
```diff
@@ -465,6 +465,24 @@ backupHandler:
...
     # targetMemoryUtilizationPercentage: 80
 
+insightsHandler:
+  enabled: true
+  replicaCount: 2
+  image:
+    repository: uselagoon/insights-handler
+    pullPolicy: Always
+    # Overrides the image tag whose default is the chart appVersion.
+    tag: ""
```
tobybellwood, line 475:
```suggestion
    repository: testlagoon/insights-handler
    pullPolicy: Always
    # Overrides the image tag whose default is the chart appVersion.
    tag: "pr-2997"
```

tobybellwood: If we do it this way, we can test that this works, and then once the corresponding Lagoon release has come out, we can revert this commit

tobybellwood:
we need to be able to control the setting for the `s3-host` to be used from the chart - same as the logs2s3 service

tobybellwood:
THe insights-handler service doesn't expose a port 3000, so the liverness and readiness checks don't work - we can either disable the checks, or expose a port (unnecessarily?)

charts/lagoon-core/templates/insights-handler.deployment.yaml
```diff
@@ -0,0 +1,112 @@
...
+        - name: RABBITMQ_PASSWORD
+          valueFrom:
+            secretKeyRef:
+              name: {{ include "lagoon-core.broker.fullname" . }}
+              key: RABBITMQ_PASSWORD
+        - name: S3_FILES_BUCKET
+          value: {{ required "A valid .Values.s3FilesBucket required!" .Values.s3FilesBucket | quote }}
+        - name: S3_FILES_HOST
+          value: {{ required "A valid .Values.s3FilesHost required!" .Values.s3FilesHost | quote }}
+        {{- with .Values.s3AccessKeyID }}
```
tobybellwood, line 52:
```suggestion
        {{- with .Values.s3FilesAccessKeyID }}
```

charts/lagoon-core/templates/insights-handler.deployment.yaml
```diff
@@ -0,0 +1,112 @@
...
+        - name: S3_FILES_ACCESS_KEY_ID
+          value: {{ . | quote }}
+        {{- end }}
+        {{- with .Values.s3SecretAccessKey }}
```
tobybellwood, line 56:
```suggestion
        {{- with .Values.s3FilesSecretAccessKey }}
```

tobybellwood:
ok - I've modified this to:
* disable insights-handler (for now) in both the linter and the test-suite
* point it to the uselagoon/insights-handler:main image
* leave it available to use downstream, by enabling it in helm values

PR #366, opened 2022-01-27
changing the broker readiness and liveness probes to use rabbit-diagn…

timclifford:


tobybellwood:
using 5672 as the liveness port is recommended by Rabbit - https://github.com/rabbitmq/cluster-operator/issues/409

We could use one of the internal commands should we need a startup probe?

PR #368, opened 2022-02-03
Add service.annotations to ssh and auth-server

### LINKED ISSUE #367: Add configurable serviceAnnotation to services
We've identified that certain loadbalancers have annotations that need to be passed (e.g. session connection timeouts for ssh)

We will add a configurable serviceAnnotation (similar to podAnnotation) that allows a chart user to pass the key and value into the annotation.

tobybellwood:
This PR adds a service.annotations option to the Helm Charts for the lagoon-core ssh and auth-server services.

These annotations can be required by service providers to allow additional configuration options.

Tested as:
```
ssh:
  service:
    annotations:
      service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: "600"
```
With the resultant service showing the annotation
```
tobybellwood@pop-os:~/sites/lagoon-main (⎈ |kind-lagoon-local:default) $ kubectl get service lagoon-core-ssh -n lagoon -o yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    meta.helm.sh/release-name: lagoon-core
    meta.helm.sh/release-namespace: lagoon
    service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: "600"
  creationTimestamp: "2022-02-03T01:23:59Z"
  labels:
```

If we need to add annotations to other services, happy to add them into this PR.

Closes #367

### DIFF
--- a/.github/workflows/test-suite.yaml
+++ b/.github/workflows/test-suite.yaml
@@ -7,6 +7,7 @@ jobs:
   test-suite:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         test:
         - active-standby-kubernetes
--- a/Makefile
+++ b/Makefile
@@ -151,7 +151,7 @@ install-mongodb:
 		bitnami/mongodb
 
 .PHONY: install-minio
-install-minio:
+install-minio: install-ingress
 	$(HELM) upgrade \
 		--install \
 		--create-namespace \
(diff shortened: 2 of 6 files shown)


tobybellwood:
I've also had to add a couple changes in to make the action more stable
* installing ingress ahead of minio to avoid a race condition
* adding fail-fast to avoid all tests being failed if a single matrix test fails - this will prove handy in multi-run scenarios

shreddedbacon (approved):
I'm pretty happy with this, it will do what is required

PR #372, opened 2022-02-09
Update lagoon-remote to use latest lagoon-build-deploy chart

shreddedbacon:


shreddedbacon:
All test have passed successfully in separate runs, failed tests appear to be because the API is not fully operation when the first test starts to run. We need to address this in core to find out why, I have a feeling it is related to the `api-db` sometimes getting hung up and crashing but not restarting properly. There are plans to move away from `api-db` in the future though, as this is already a problematic point.

PR #377, opened 2022-02-14
Feature/insights remote

bomoko:


tobybellwood:
Attached is a payload collected from the broker (ignore the `lagoon.sh/insights` label - it now collects on `lagoon.sh/insightsType` instea
[broker.txt](https://github.com/uselagoon/lagoon-charts/files/8058720/broker.txt)
d

PR #386, opened 2022-02-18
Adding insights feature flag env vars to lagoon-remote chart

timclifford:


shreddedbacon:
Ahh you need to bump the chart and application version (once the `remote-controller` version is released)
https://github.com/uselagoon/lagoon-charts/blob/main/charts/lagoon-build-deploy/Chart.yaml#L18-L20

PR #387, opened 2022-02-21
Use Keycloak startup readiness signalling

smlx:
This PR uses the startup readiness signalling implemented in https://github.com/uselagoon/lagoon/pull/3042 to ensure the Keycloak pod does not become `Ready` in the k8s API until its configuration is complete.


smlx:
As discussed, lets hold off on merging this until a new Lagoon release containing https://github.com/uselagoon/lagoon/pull/3042 is cut.

PR #395, opened 2022-02-28
Update lagoon-charts components to support Kubernetes 1.22

tobybellwood:
This PR contains all the changes needed to get Lagoon running on Kubernetes 1.22. As the necessary updates are made to the various components, the overrides in here will no longer be needed and will eventually be removed.

Anybody wishing to deploy Lagoon on Kubernetes **MAY** be able to use the configuration herein to get it working, but be warned that it is still in a changing state, so dragons etc etc.


tobybellwood:
work completed elsewhere

PR #399, opened 2022-03-03
Updates lagoon-remote chart depdendencies

bomoko:
This simply updates the dependencies for insights remote and build-deploy

### DIFF
--- a/charts/lagoon-remote/values.yaml
+++ b/charts/lagoon-remote/values.yaml
@@ -118,7 +118,6 @@ lagoon-build-deploy:
   # taskSSHHost: ""
   # taskSSHPort: ""
   # taskAPIHost: ""
-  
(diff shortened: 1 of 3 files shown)


tobybellwood:
We should add the new possible values for insights-remote (and build-deploy) into the lagoon-remote values file as examples
https://github.com/uselagoon/lagoon-charts/compare/lagoon-build-deploy-0.8.2...lagoon-build-deploy-0.8.5

charts/lagoon-remote/values.yaml
```diff
@@ -118,6 +118,9 @@ lagoon-build-deploy:
   # taskSSHHost: ""
   # taskSSHPort: ""
   # taskAPIHost: ""
+
```
tobybellwood, line 121:
```suggestion

```

PR #403, opened 2022-03-14
fix: bump the keycloak startup probe timeout to 15m

smlx:
Keycloak sometimes takes >10 minutes to start - I've seen Keycloak get killed before completing configuration both locally and in lagoon-charts CI.


tobybellwood (approved):
yup - I've seen it perilously close to 10min too, no harm.

PR #410, opened 2022-03-16
run multiple tests per cluster

tobybellwood:


smlx:
I tried this a while back but ran into problems including:
* Not enough RAM on the CI boxes (random errors and test timeouts)
* Not enough disk space for images on the CI boxes (random errors and test timeouts; that's what https://github.com/uselagoon/lagoon-charts/blob/main/.github/workflows/test-suite.yaml#L146-L150 is supposed to help fix)
* some tests have duplicate ingress names/domains which can cause spurious failures when test suites running in parallel try to create the same ingress

Just a few things to keep an eye out for :slightly_smiling_face:

tobybellwood:
ah yeah - just seeing what the options are, esp now that we can re-run indiv matrix jobs

tobybellwood:
not a priority now

PR #412, opened 2022-03-23
Minor updates to insights-core templates to match others

shreddedbacon:

### DIFF
--- a/charts/lagoon-core/templates/insights-handler.deployment.yaml
+++ b/charts/lagoon-core/templates/insights-handler.deployment.yaml
@@ -53,6 +53,16 @@ spec:
         - name: S3_FILES_REGION
           value: {{ . | quote }}
         {{- end }}
+        - name: S3_FILES_ACCESS_KEY_ID
+          valueFrom:
+            secretKeyRef:
+              name: {{ include "lagoon-core.api.fullname" . }}
+              key: S3_FILES_ACCESS_KEY_ID
+        - name: S3_FILES_SECRET_ACCESS_KEY
+          valueFrom:
+            secretKeyRef:
+              name: {{ include "lagoon-core.api.fullname" . }}
+              key: S3_FILES_SECRET_ACCESS_KEY
         - name: JWT_SECRET
           valueFrom:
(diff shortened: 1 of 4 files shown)


tobybellwood:
we should add in the `S3_USESSL` variable here too

shreddedbacon:
> we should add in the `S3_USESSL` variable here too

https://github.com/uselagoon/lagoon-charts/blob/main/charts/lagoon-core/templates/insights-handler.deployment.yaml#L75
We don't need to.
I changed the logic to be enabled by default in the insights handler, and if a user wanted to disable it they can do so via the additionalEnvs values

charts/lagoon-core/templates/insights-handler.deployment.yaml
```diff
@@ -49,14 +49,6 @@ spec:
...
         - name: S3_FILES_HOST
           value: {{ required "A valid .Values.s3FilesHost required!" .Values.s3FilesHost | quote }}
-        {{- with .Values.s3FilesAccessKeyID }}
-        - name: S3_FILES_ACCESS_KEY_ID
-          value: {{ . | quote }}
-        {{- end }}
-        {{- with .Values.s3FilesSecretAccessKey }}
-        - name: S3_FILES_SECRET_ACCESS_KEY
-          value: {{ . | quote }}
-        {{- end }}
```
smlx, line 59: Could we specify the individual env vars here instead of importing all those in the api secret? Reasons are:
* makes it easy to see at a glance which variables the pod actually uses
* sensitive info that the pod doesn't need is kept out of the pod's environment

What I mean is to do something like this instead: https://github.com/uselagoon/lagoon-charts/blob/aa9ce769295844b07b081fefe602e3e45c355105/charts/lagoon-core/templates/api-redis.deployment.yaml#L33-L38

smlx (changes requested):

charts/lagoon-core/templates/insights-handler.deployment.yaml
shreddedbacon: Sure

PR #417, opened 2022-04-05
Update Lagoon charts for Kubernetes 1.22 (min version now 1.19)

tobybellwood:
This PR updates the ingresses, hpas and priorityclasses in preparation for Kubernetes 1.22. As these changes are only compatible with Kubernetes 1.19 onwards, the kubeVersion constraint has been added to the chart.yaml

In addition, the lagoon-gatekeeper chart release 0.4.0 has been marked as deprecated, and only compatible with <1.22

This PR also disables linting for the lagoon-gatekeeper chart and lagoon-insights-remote (as it needs remediation to start successfully without a broker present (https://github.com/uselagoon/insights-remote/issues/2)

Note that the ingresses and priorityclasses have already been auto-patched to their GA versions by Kubernetes on install, so this is mostly a tidyup to remove the deprecation notices.

In this PR, I have only migrated lagoon-core to a major version, as that is the primary chart for Lagoon (and also where the most change is occurring). We can mark Lagoon-remote as 1.0 after the dependent charts are updated,


charts/lagoon-core/templates/broker.hpa.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
 spec:
   scaleTargetRef:
     apiVersion: apps/v1
-    kind: Deployment
+    kind: StatefulSet
```
smlx, line 11: Nice catch

smlx (approved):
The chart changes LGTM. 1.0.0 :tada:

.github/workflows/lint-test.yaml
```diff
@@ -60,6 +62,15 @@ jobs:
...
   lint-test:
     runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.experimental }}
+    strategy:
+      fail-fast: false
+      matrix:
+        kindest_node_version: [v1.21.1, v1.22.7]
+        experimental: [false]
+        include:
+          - kindest_node_version: v1.23.4
```
smlx, line 72: Indentation doesn't match the rest of the file here.

.github/workflows/lint-test.yaml
```diff
@@ -60,6 +62,15 @@ jobs:
   # default.ct.yaml
   lint-test:
     runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.experimental }}
```
smlx, line 65: It would be good to keep the main lint-test simple and straightforward. E.g. on the latest stable kind version. Then maybe we can have a separate PR tag `experimental` which triggers an experimental workflow?

Doing it this way with boolean matrix values seems to be putting many concepts into a single workflow which makes it difficult to follow.

smlx (changes requested):

.github/workflows/lint-test.yaml
smlx: To elaborate a bit, what this workflow was intended to do is `lint-test` the chart via chart-testing on the stable kind version. This should pick up problems with chart templates, runs on every PR, and is required to merge.

The changes that are being added here are backwards and forwards compatibility changes for various kind versions. I think those should go into a separate workflow which can have experimental flags etc. and probably wouldn't be required to go green before merging.

What do you think?

tobybellwood: fixed up - moved matrix tests into different workflow that we can mark as non-required.

smlx (approved):
LGTM

PR #420, opened 2022-04-08
fix: refactor keycloak admin user variable

smlx:
This changes the way the keycloak admin user is added to simplify it and
align it with the way we handle default non-sensitive values in the rest
of the chart.


smlx:
Sorry @tobybellwood I messed up the review on the PR yesterday. This is the way we do non-sensitive variables elsewhere in the chart so I thought to try to keep things consistent and easier to understand. Does this change make sense to you?

tobybellwood:
yup - looks ok to me!

PR #425, opened 2022-04-19
update logging operator dependency to 3.17.4

tobybellwood:
the 3.17.3 release of the logging-operator added HostTailer and EventTailer, enabled by default. If the logging operator is installed without the updated CRDs, it will crashloop.

This PR updates the logging-operator to 3.17.4 where they are [disabled by default](https://github.com/banzaicloud/logging-operator/pull/958), whilst we work out how to automate the update of the CRDs after an update

and yes, we could do this, but I just don't know if we should, or how, or where
```
helm show crds banzaicloud-stable/logging-operator --version 3.17.3 | kubectl apply -f -
```

### DIFF
--- a/charts/lagoon-logging/Chart.lock
+++ b/charts/lagoon-logging/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: logging-operator
   repository: https://kubernetes-charts.banzaicloud.com
-  version: 3.17.3
-digest: sha256:3fcd9f977a737ea57bf18a4ffa2e587146f1ff87400c99b5dfcd134a5cdb8321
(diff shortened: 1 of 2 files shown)


smlx:
Thanks @tobybellwood I did put upgrade notes in the version that introduced the breaking change, but maybe it didn't have enough visibility https://github.com/uselagoon/lagoon-charts/releases/tag/lagoon-logging-0.59.0

tobybellwood:
yah - I did miss those :fearful: - the releases in this repo are thick and fast!

I think we need to look how to handle these crd updates along the way! I updated them in our test cluster (test1), but having them disabled for now is also easy!

PR #437, opened 2022-06-02
Update upstream helm charts in makefile/CI

tobybellwood:
Bitnami has deprecated a number of the helm charts that we were using in CI.

This PR updates the following:
* `bitnami/mariadb` from 9.3.13 to 10.5.1 (latest MariaDB 10.5.x)
* `bitnami/postgresql` from 10.4.8 to 10.16.2 (latest PostgreSQL 11.x)
* `bitnami/mongodb` from 10.16.4 to 11.2.0 (latest MongoDB 4.x)
* `bitnami/minio` from 8.1.9 to 11.6.3 (latest Minio 2022.5.x)
  * It also updates the values in the chart for the user/pass

In addition, this PR updates:
* `ingress-nginx/ingress-nginx` from 3.31.1 to 4.1.3 (latest ingress-nginx 1.2.x)
  * it also adds the correct IngressClass annotations for Lagoon
* `harbor/harbor` from 1.5.5 to 1.9.1 (latest Harbor 2.5.x)
  * In order to support Harbor >2.2, the chart has also been configured to have Harbor controlled by the remote-controller

I've also added the `dbaas` test suite, that ensures that the dbaas-operator, and more complex build nuances are working - it has the advantage of building drupal and nodejs sites inside the single routine)


charts/lagoon-build-deploy/ci/linter-values.yaml
```diff
@@ -5,3 +5,10 @@ lagoonTargetName: ci-local-control-k8s
 taskSSHHost: lagoon-core-ssh.lagoon.svc
 taskSSHPort: 2020
 taskAPIHost: http://lagoon-core-api.lagoon.svc:80
+extraArgs:
```
shreddedbacon, line 8: probably dont need to define these here really, the linter test doesnt care if harbor is enabled or not

tobybellwood: I've only added them here in case people look to the linter for a template :wink:

Issue #443, opened 2022-06-28
Add startup probes to stateful services


smlx:
We have the following services which can sometimes fail liveness probes and be killed by kubernetes before having a chance to start up properly:
* `api-db`
* `keycloak-db`
* `broker` (as per #424)

This can be bad because a) it can cause state/data corruption, and b) it is often slower than just waiting for the startup to complete.

We should add startup probes to these services and allow a significantly longer period for the startup to complete compared to the regular liveness probes.

Issue #424, opened 2022-04-19
Increase liveness/readiness for broker


shreddedbacon:
Currently the liveness/readiness checks are 10 seconds. This should be increased, have experienced recently that the broker pod failed to start and would crash after 10 seconds but increasing this to 60s gave the pod enough time to start up. Assuming the size of the messages to process on disk needed more time for broker to start.

PR #445, opened 2022-06-30
Startup probes

smlx:
This change just hard-codes a 10 minute maximum startup timeout on the stateful pods `api-db`, `keycloak-db`, and `broker`.

Is it okay being hard-coded? Or should this be configurable? I'm thinking YAGNI, but there might be a use case.

Closes: #424
Closes: #443


shreddedbacon (commented):
I know hard coding stuff has bitten us in the past, so I'm on the fence about it.

smlx:
Ok, have made startup probes a config option. cc @shreddedbacon

shreddedbacon (approved):

Issue #444, opened 2022-06-28
Disable the PodSecurityPolicy on the logging-operator chart by default


smlx:
The logging operator chart, which is used by lagoon-logging, enables a pod security policy by default. This will not install into a Kubernetes 1.22+ cluster, so we should disable it manually.

smlx:
Oops my mistake. The PSP API will not be removed until k8s 1.25. So hopefully upstream will remove it and we can just bump the dependency.

PR #446, opened 2022-07-01
Disable PSP in logging-operator

smlx:
The PSP API is deprecated in recent k8s versions, so we need to disable it.

Closes: #444


smlx:
API is not yet removed in the latest k8s 1.24, so we don't need to do this yet.

PR #448, opened 2022-07-11
update charts to use the activestandby task image overrides

shreddedbacon:
Lagoon-core uses an old method for calculating the active-standby task image, rather than one that is used during the build/test phase. This PR updates the charts to support providing the image built as part of tests so we can properly use active/standby images in tests.


smlx:
This PR removes the kubectl-build-deploy-dind override functionality. Is that intentional?

shreddedbacon:
> This PR removes the kubectl-build-deploy-dind override functionality. Is that intentional?

Yes, this functionality in lagoon-core has been removed for some time (since remote-controller became the default) and the image is no longer defined in the API only in the remotes.

smlx:
Ok, in that case this might be a major version bump on the chart because it is removing an API from the values file.

PR #451, opened 2022-07-15
Bump keycloak startup probe

smlx:
Keycloak can take longer than 15 minutes to start from scratch in CI.


shreddedbacon (commented):
I don't see any downsides to this

PR #454, opened 2022-07-23
add ingressClassName to Lagoon-core Ingresses

tobybellwood:
This PR does the minimum needed to add the IngressClassName field to the lagoon-core ingresses to allow Lagoon-core to deploy into clusters with multiple ingresses

It is configured from a Values.yaml setting: `defaultIngressClass: nginx`, and can be overridden per ingress with a coalesce
```
defaultIngressClass: nginx
...
webhookHandler:
  ingress:
    ingressClassName: other-nginx
```
I've deliberately chosen to call it defaultIngressClass to avoid confusion with defaultIngress (if set) - even though in practice, the two may be named similarly

I've also updated the annotations commented out in the values file to remove the deprecated `kubernetes.io/ingress.class: nginx` setting, and added in the cert-manager default as an example. I did an initial pass at merging/overwriting a default annotation, but it got ugly quickly.

I've refrained from adding the cert-manager annotation in as a default to all ingresses as :shrug: not everyone might want it?

Also discovered a surprise ingress in logs-concentrator that I didn't know (or had forgotten) about


charts/lagoon-logs-concentrator/ci/linter-values.yaml
```diff
@@ -2,6 +2,7 @@
 elasticsearchHost: "logs-db-service.elasticsearch.svc.cluster.local"
 elasticsearchAdminUser: "admin"
 elasticsearchAdminPassword: "securepass"
+defaultIngressClass: nginx
```
smlx, line 5: minor nit: should we make this `defaultIngressClassName` for consistency with the other option under each ingress?

charts/lagoon-core/ci/linter-values.yaml
```diff
@@ -15,6 +15,7 @@ s3BAASAccessKeyID: baas
 s3BAASSecretAccessKey: baas123
 lagoonUIURL: http://ui:9101
 lagoonWebhookURL: http://webhook:11213
+defaultIngressClass: nginx
```
smlx, line 18: This doesn't have any effect since no ingress objects are enabled AFAIK.

charts/lagoon-core/templates/api.ingress.yaml
```diff
@@ -12,6 +12,9 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
+  {{- if or (.Values.api.ingress.ingressClassName) (.Values.defaultIngressClass) }}
+  ingressClassName: {{coalesce .Values.api.ingress.ingressClassName .Values.defaultIngressClass }}
+  {{- end }}
```
smlx, line 17: It might be clearer to make this logic look like this:
```
  {{- with coalesce .Values.api.ingress.ingressClassName .Values.defaultIngressClass }}
  ingressClassName: {{ . }}
  {{- end }}
```

smlx (changes requested):
I see that you've updated the comments in `values.yaml` to indicate how to set the cert-manager annotation, but could we automatically annotate the ingress objects if the ingress class is set?

tobybellwood:
That was my first thought, but died away from it in case people aren't using cert-manager - and something else instead (not that it would likely interfere though). Could we assume it's so dominant that we set them?

smlx:
Unknown annotations will just be ignored by other services AFAIK, so I think it's pretty safe to add them.

tobybellwood:
I'll put them back in 🥸

charts/lagoon-core/templates/api.ingress.yaml
tobybellwood: The downside of this is b/c - if there is no ingressClassName or defaultIngressClassName specified, it outputs a null field on the Ingress - which may be undesirable?

charts/lagoon-core/ci/linter-values.yaml
tobybellwood: yah, I've left it in, just to ensure that someone looking copies it :wink:

charts/lagoon-logs-concentrator/ci/linter-values.yaml
tobybellwood: fixed

charts/lagoon-core/templates/api.ingress.yaml
```diff
@@ -7,11 +7,17 @@ metadata:
   name: {{ $fullName }}
   labels:
     {{- include "lagoon-core.api.labels" . | nindent 4 }}
-  {{- with .Values.api.ingress.annotations }}
   annotations:
```
smlx, line 10: I think it would be nice to avoid having this empty by wrapping it in a conditional, or at least enabling an ingress in CI so it is tested in all the k8s versions. Not a showstopper however.

charts/lagoon-core/ci/linter-values.yaml
```diff
@@ -15,6 +15,7 @@ s3BAASAccessKeyID: baas
 s3BAASSecretAccessKey: baas123
 lagoonUIURL: http://ui:9101
 lagoonWebhookURL: http://webhook:11213
+defaultIngressClassName: nginx
```
smlx, line 18: IMO this should either be removed or one of the ingresses should be enabled. As it is, there is no point having this value here because it has no effect and could cause confusion later.

smlx (approved):
Couple of minor nits but we can go and fix those up later if required.

I also realised that the ingress in the logs-concentrator is unused and shouldn't be there! I'll send a PR to remove it :grimacing:

PR #459, opened 2022-07-28
Update lagoon-remote to use latest lagoon-build-deploy-0.14.0

shreddedbacon:
Also added a post installation note about updating CRDs


shreddedbacon:
Do we want to display the Lagoon logo like we do with core when installing remote? Might make it more noticeable that there is an action item once installed/updated?

tobybellwood:
> Might make it more noticeable that there is an action item once installed/updated?

Maybe just an `action required` callout? although I do like the idea of a lagoon core and lagoon remote logo...

shreddedbacon:
> > Might make it more noticeable that there is an action item once installed/updated?
> 
> Maybe just an `action required` callout? although I do like the idea of a lagoon core and lagoon remote logo...

Already a message that says "ensure you do x" so covered there. Was more just making it more obvious by showing the logo above it like we do with core.

PR #467, opened 2022-08-16
test with updated mariadb images

tobybellwood:


tobybellwood:
not proceeding with this direction, testing only

PR #476, opened 2022-09-08
Improve probes on api-db pod

smlx:
This is intended to help automatically recover from the case where the api-db pod becomes unresponsive.


tobybellwood (approved):
Yup, seems to address the issue, and is easily configurable

PR #479, opened 2022-09-09
chore: run tmate first

smlx:
This change reduces the timeout to 1m and runs tmate first to make it easier to catch a shell.

It also removes the old api-db checks since that was fixed in #476.


shreddedbacon (approved):

tobybellwood (approved):
yup - happy with this!

PR #480, opened 2022-09-12
Drop logs-tee

smlx:
This feature of the chart was used only for migrating from the old Lagoon logging system which is long since deprecated.


tobybellwood (approved):
LGTM

PR #482, opened 2022-09-12
Repoint lagoon-service-images to default "latest"

tobybellwood:


smlx:
Can we keep a version pin in these by default? This is important for reliability because e.g. `logs-concentrator` sometimes has particular image version requirements for compatibility with the various versions of Elasticsearch/Opensearch that we support. If we have these pods auto-upgrade as they restart depending on where the `latest` tag points it's going to get messy. `logs-dispatcher` has similar requirements.

tobybellwood:
yes - happy to likely code the versions into the values files once I've got the first releases out. These won't move very often, so it shouldn't be too hard.

I also want to investigate auto-bumps to these helm values (via renovate or dependabot) to do this.

PR #487, opened 2022-09-21
Container memory limit

smlx:
This change supports container memory limits for containers deployed by Lagoon.

See https://github.com/uselagoon/build-deploy-tool/pull/120 for the other part of this feature.


smlx:
I gave this flag an `ADMIN_` prefix to distinguish it as only being set by administrators (via helm chart values) and never by users.

tobybellwood (approved):
LGTM

PR #493, opened 2022-10-07
Remove old kubernetes-build-deploy serviceaccount from lagoon-remote

### LINKED ISSUE #294: Remove kubernetes-build-deploy objects from lagoon-remote
These are no longer used in Lagoon 2.0 :tada:

shreddedbacon:
`lagoon-build-deploy` creates its own service account now, and this service account should be the one used in the Lagoon API.

This is possibly a breaking change (removing a cluster token), in which Lagoon operators may need to update their Lagoon API `openshift/kubernetes` token for the respective clusters to use the token provided by the `lagoon-build-deploy` service account if they are not already using this.

closes #294


shreddedbacon:
Now that #608 has been merged, and is the current method for long lived tokens to use for SSH in lagoon-core, this PR should merge so we can remove the old references

tobybellwood:
merging in #627

Issue #496, opened 2022-10-18
HorizontalPodAutoscaler update API version to autoscaling/v2


britdm:
Requesting an update to HPA API version references in `charts/lagoon-core/templates` from `autoscaling/v2beta1` to `autoscaling/v2` to move to a supported API version for v1.22+.

shreddedbacon:
Originally closed in #542

PR #497, opened 2022-10-21
Bump lagoon-core HPAs to autoscaling/v2

smlx:
This change bumps the autoscaling API used by the `lagoon-core` HPAs to `autoscaling/v2`.

This is a breaking change because it means the chart no longer supports kubernetes < 1.23.

https://kubernetes.io/docs/reference/using-api/deprecation-guide/#horizontalpodautoscaler-v126

Closes: #496


tobybellwood:
closed in #542

PR #498, opened 2022-10-21
Lagoon logs concentrator hpa v2

smlx:
This change bumps the autoscaling API used by the `lagoon-logs-concentrator` HPA to `autoscaling/v2`.

This is a breaking change because it means the chart no longer supports kubernetes < 1.23.

https://kubernetes.io/docs/reference/using-api/deprecation-guide/#horizontalpodautoscaler-v126


tobybellwood:
closed in #542

PR #499, opened 2022-10-24
feat: add storage-calculator to remote

shreddedbacon:
Adds support for running the storage-calculator (https://github.com/uselagoon/storage-calculator) in lagoon-remote.

This calculator communicates with lagoon-core via rabbitmq. To be able to use this, the minimum version of lagoon-core that supports remote storage-calculator via rabbitmq is v2.10.0 (https://github.com/uselagoon/lagoon/releases/tag/v2.10.0 see note under `Service consolidation`)


tobybellwood:
@shreddedbacon - should we be adding an `enabled:` field under it in the remote - i could imagine disabling it would be beneficial to some users?

shreddedbacon:
> @shreddedbacon - should we be adding an `enabled:` field under it in the remote - i could imagine disabling it would be beneficial to some users?

Yes, good idea :smile:

tobybellwood (approved):
LGTM

tobybellwood:
ooh, actually @shreddedbacon - do you want to add the [annotations](https://artifacthub.io/docs/topics/annotations/helm/#supported-annotations) for this one to the Chart.yaml?

```
  artifacthub.io/changes: |
    - kind: added
      description: Support for storage-calculator, disabled by default
```

PR #502, opened 2022-11-01
chore: use the new syntax for output variables in github actions

smlx:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


tobybellwood (approved):
seems legit :wink:

PR #504, opened 2022-11-03
Bump ssh-portal version in lagoon-remote and lagoon-core

smlx:
This is a minor update to the ssh-portal which switches JWT parsing/validation libraries to a better maintained one. See https://github.com/uselagoon/lagoon-ssh-portal/pull/126.


tobybellwood (changes requested):
Can you also add/replace the `artifacthub.io/changes` annotations in the chart.yaml files? There is an example in lagoon-remote - or fuller update in #503

smlx:
sure thing, will do. I've just noticed something else I need to fix, so will update the PR once it is ready.

tobybellwood (approved):

PR #511, opened 2022-11-11
chore: bump lagoon-build-deploy version

shreddedbacon:
This bumps the `lagoon-build-deploy` version to the latest `v0.17.0` chart version.

### DIFF
--- a/charts/lagoon-remote/Chart.lock
+++ b/charts/lagoon-remote/Chart.lock
@@ -14,5 +14,5 @@ dependencies:
 - name: nats
   repository: https://nats-io.github.io/k8s/helm/charts/
   version: 0.18.2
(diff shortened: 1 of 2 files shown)


shreddedbacon:
One question about this, the `nats` chart also bumped a patch version here, which I'm hoping is fine. Otherwise should we look at pinning `nats` to be a specific version? Not entirely sure about this one yet.

tobybellwood:
Maybe we should merge https://github.com/uselagoon/lagoon-charts/pull/508 ahead of this then - it updates both NATS charts at once

shreddedbacon:
> Maybe we should merge #508 ahead of this then - it updates both NATS charts at once

SGTM

PR #513, opened 2022-11-17
Spot logging

smlx:
Add support for logs collection from spot instances.


smlx:
The test failure is due to https://github.com/banzaicloud/logging-operator/issues/1077

I'll send a PR upstream to fix it.

tobybellwood:
ugh, we should have got the artifacthub annotations in here :cry:

smlx:
oh no! sorry Toby, maybe we need a CI check for that?

PR #514, opened 2022-11-18
Add ssh-token service

smlx:
lagoon-core:
* Add the [ssh-token](https://github.com/uselagoon/lagoon-ssh-portal#ssh-token) service. This service is disabled by default to avoid an upgrade spinning up a new load balancer on chart upgrade, but it should probably be enabled by default eventually?
* Allow disabling the ssh and auth-server services via `ssh.enabled`.
* Update ssh-portal-api to match ssh-token version.

lagoon-remote:
* Update ssh-portal version to match the other ssh services.

lagoon-test:
* Add ssh-token environment variables to test container.

### DIFF
--- a/charts/lagoon-core/Chart.yaml
+++ b/charts/lagoon-core/Chart.yaml
@@ -21,7 +21,7 @@ type: application
 # time you make changes to the chart and its templates, including the app
 # version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.18.0
+version: 1.19.0
 
 # This is the version number of the application being deployed. This version
 # number should be incremented each time you make changes to the application.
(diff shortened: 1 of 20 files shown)


smlx:
This PR doesn't enable `ssh-token` or disable `ssh`, but testing with `ssh-token` instead of `ssh` does [pass the Lagoon test suite](https://ci.lagoon.sh/blue/organizations/jenkins/lagoon/detail/ssh-token/23/pipeline/60).

shreddedbacon:
If the ssh-token service is enabled, and ssh.enabled is set. You will have two services that can generate tokens, is this correct?

smlx:
> If the ssh-token service is enabled, and ssh.enabled is set. You will have two services that can generate tokens, is this correct?

Yes, correct.

smlx:
This PR is tested in https://github.com/uselagoon/lagoon/pull/3355 and now passes all tests in Jenkins except the `rsh` tests.

smlx:
This PR now passes the entire Lagoon test suite. See https://github.com/uselagoon/lagoon/pull/3355/

tobybellwood (approved):
ok. Safe to add a new service disabled by default.

PR #520, opened 2022-11-23
Lagoon logging tests

smlx:
Add a test suite for lagoon-logging, and bump the logs-dispatcher image for better compatibility with recent k8s versions.

See https://github.com/uselagoon/lagoon-service-images/pull/10

PR #524, opened 2022-11-28
Bump logging-operator dependency

smlx:
Bump the logging-operator dependency to the latest version which adds support for k8s 1.25+.


smlx:
Bump two minor versions since this should go in after #520

tobybellwood (approved):
LGTM

PR #527, opened 2022-12-06
Remove old lagoon-gatekeeper install instructions

rocketeerbkw:
The need to use `--set=constraints.create=true` was removed in https://github.com/uselagoon/lagoon-charts/commit/45499092472b472250b0afddf41562f6f45b138a.


smlx:
This entire chart is deprecated: https://github.com/uselagoon/lagoon-charts/blob/main/charts/lagoon-gatekeeper/Chart.yaml#L13

Probably we should just remove it entirely. Thoughts @tobybellwood ?

PR #526, opened 2022-11-29
Consolidated Lagoon 2.12 release - Kubernetes 1.21 minimum

CGoodwin90:
Removed unused tests & included Services
Updates minimum supported Kubernetes version for all charts to 1.21

Updated for lagoon core release https://github.com/uselagoon/lagoon/releases/tag/v2.12.0

PR #531, opened 2022-12-13
Update api-data-watcher-pusher startup probe

smlx:
Update the api-data-watcher-pusher startup probe for the new readiness file added in https://github.com/uselagoon/lagoon/pull/3348

This is a draft because the linked PR won't make into a Lagoon release until v2.12.0, at which point this PR can be rebased.


tobybellwood:
closing and adding to #526 tied to lagoon release

PR #535, opened 2022-12-23
Added annotations support to ssh-portal service

### LINKED ISSUE #534: add annotation support to ssh-portal service in lagoon-remote
FOllowing from https://github.com/uselagoon/lagoon-charts/pull/368 - now we have ssh-portal support in lagoon-remote, any AWS implementations will need an annotation to extend the timeout for the loadbalancer

tobybellwood:
closes #534

### DIFF
--- a/charts/lagoon-remote/Chart.yaml
+++ b/charts/lagoon-remote/Chart.yaml
@@ -19,7 +19,7 @@ type: application
 # This is the chart version. This version number should be incremented each
 # time you make changes to the chart and its templates, including the app
 # version.
-version: 0.70.0
+version: 0.70.1
 
 dependencies:
 - name: lagoon-build-deploy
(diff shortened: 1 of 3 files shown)


charts/lagoon-remote/Chart.yaml
```diff
@@ -19,7 +19,7 @@ type: application
 # This is the chart version. This version number should be incremented each
 # time you make changes to the chart and its templates, including the app
 # version.
-version: 0.70.0
```
smlx, line 22: I think this should be 0.71.0, as its a new feature - not a bugfix.

smlx (changes requested):
Minor versioning nitpick but otherwise LGTM!

charts/lagoon-remote/Chart.yaml
tobybellwood: ok - fixed

PR #536, opened 2022-12-23
Bump ssh-portal, ssh-portal-api, and ssh-token service versions

smlx:
Upgrade to the latest versions which fix https://github.com/uselagoon/lagoon-ssh-portal/issues/158


smlx:
This should go in after https://github.com/uselagoon/lagoon-charts/pull/535

tobybellwood (approved):
all matching and correct.

PR #540, opened 2023-01-05
Switch logs-concentrator default configuration from Elasticsearch to Opensearch

smlx:
Opensearch and Elasticsearch have now diverged enough that there are separate fluentd plugins for each project.

The old fluentd Elasticsearch plugin doesn't work anymore with Opensearch v2, so this PR switches the default logs-concentrator image to a version which includes the fluentd Opensearch plugin (which supports Opensearch v2), and updates the default fluentd configuration to assume Opensearch log storage rather than Elasticsearch.

Note that this is a breaking change that requires users to update their values files.


tobybellwood (approved):
Looks OK to me - we should call out the variable rename in the release notes though

PR #542, opened 2023-01-05
configured dynamic apiVersion for HorizontalPodAutoscaler

tobybellwood:
There are two different versions of the autoscaling API in use for HorizontalPodAutoscalers

- autoscaling/v2beta1 is being deprecated in Kubernetes 1.25
- autoscaling/v2beta2 is being deprecated in Kubernetes 1.26 (and required a small change from v2beta1
- autoscaling/v2 has been available since Kubernetes 1.23

This PR adds a helper step to determine whether the autoscaling/v2 API is available, and Kubernetes is 1.23 - and then sets autoscaling/v2, otherwise faills back to the autoscaling/v2beta2 API. It also updates the spec to match the beta2 fields.

In addition, it enables the HPA for the backup-handler service, and adds Kubernetes 1.26 to the matrix to ensure all versions are tested ok.


shreddedbacon (commented):
I'm not really placed to review this properly. But it looks ok to me from quick :eyes:.

One thing is, is the metrics-server actually required? It seems like it would be if you wanted to check that HPAs work, but none of our tests check this functionality. Installing metrics-server just seems like extra time for no real gain in the test-suite.

.github/workflows/lint-test-matrix.yaml
```diff
@@ -48,5 +49,10 @@ jobs:
         (steps.list-changed.outputs.changed == 'true') ||
         (contains(github.event.pull_request.labels.*.name, 'needs-testing'))
 
+    - name: Install metrics-server
+      id: metrics-server
+      run: |
+        helm upgrade --install --set args={--kubelet-insecure-tls} metrics-server metrics-server/metrics-server --namespace kube-system
+
```
smlx, line 56: Do we actually need to install the metrics server? If it isn't installed the API will still accept the HPA objects, they just won't do anything.

Having the metrics server installed will slow down tests, may cause errors we don't actually care about if something goes wrong with the metrics server, and also may cause scaling to occur in Lagoon components in CI which will use more resources etc. I'm not sure if any of that is desirable?

charts/lagoon-logs-concentrator/templates/_helpers.tpl
```diff
@@ -61,3 +61,14 @@ Create the name of the service account to use
 {{- default "default" .Values.serviceAccount.name }}
 {{- end }}
 {{- end }}
+
+{{/*
+Get HorizontalPodAutoscaler API Version
+*/}}
+{{- define "lagoon-logs-concentrator.hpa.apiVersion" -}}
+  {{- if and (.Capabilities.APIVersions.Has "autoscaling/v2") (semverCompare ">= 1.23-0" .Capabilities.KubeVersion.Version) -}}
```
smlx, line 69: Why do we need two conditional checks here? If `autoscaling/v2` is available, I would have thought the version of k8s is irrelevant?

smlx (commented):
I have a couple of questions about this change - maybe we can get away with smaller scope on this PR?

smlx:
> The reason I installed the metrics-server was to check the process for the autoscaling when enabled (the backup-handler is set to autoscale).

IMO this is out of scope for the charts CI test suite. We only care if the right objects are being created in the API for a given k8s version, and we don't care if one particular metrics server implementation is functional or not.

charts/lagoon-logs-concentrator/templates/_helpers.tpl
smlx: in that case maybe just a comment will do?

.github/workflows/lint-test-matrix.yaml
tobybellwood: I've removed it now!

charts/lagoon-logs-concentrator/templates/_helpers.tpl
tobybellwood: fixed

charts/lagoon-core/templates/_helpers.tpl
```diff
@@ -673,3 +673,14 @@ app.kubernetes.io/name: {{ include "lagoon-core.name" . }}
...
+{{/*
+Get HorizontalPodAutoscaler API Version - can be removed once Kubernetes 1.23 is the minimum
+*/}}
+{{- define "lagoon-core.hpa.apiVersion" -}}
+  {{- if (.Capabilities.APIVersions.Has "autoscaling/v2") -}}
+    {{- print "autoscaling/v2" -}}
+  {{- else -}}
+    {{- print "autoscaling/v2beta2" -}}
+  {{- end -}}
+{{- end -}}
```
smlx, line 686:
```suggestion
{{- define "lagoon-core.hpa.apiVersion" -}}
  {{- if (.Capabilities.APIVersions.Has "autoscaling/v2") -}}
    autoscaling/v2
  {{- else -}}
    autoscaling/v2beta2
  {{- end -}}
{{- end }}
```

smlx: This might be a bit simpler - same thing goes for the other helper in lagoon-logs-concentrator.

smlx (changes requested):

smlx (approved):
LGTM

PR #549, opened 2023-02-28
fix: disable HSTS by default when installing ingress-nginx via makefile

rocketeerbkw:
A minor fix to the makefile to disable HSTS headers by default in `ingress-nginx`. Needed for tests to pass in https://github.com/uselagoon/lagoon/pull/3402.


shreddedbacon (approved):
Makes sense to me

PR #552, opened 2023-03-13
chore: use rollingupdate for api for 1 at a time rollouts

shreddedbacon:
When updating the API pods when a knex migration is called, the initContainer can cause the API to fail when multiple pods attempt the migrations by locking the migration tables.

This changes the rolloutstrategy to only do 1 pod at a time, this means that only 1 initContainer should ever try to perform migrations.

### DIFF
--- a/charts/lagoon-core/Chart.yaml
+++ b/charts/lagoon-core/Chart.yaml
@@ -21,7 +21,7 @@ type: application
 # time you make changes to the chart and its templates, including the app
 # version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.25.0
+version: 1.25.1
 
 # This is the version number of the application being deployed. This version
 # number should be incremented each time you make changes to the application.
@@ -41,4 +41,4 @@ dependencies:
 annotations:
   artifacthub.io/changes: |
     - kind: changed
-      description: update Lagoon appVersion to v2.13.0
+      description: add rollout strategy for api init containers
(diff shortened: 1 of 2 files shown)


rocketeerbkw:
Since this replaces pods one at a time, it's possible that some pods are available to accept api requests but don't know about the db schema changes yet. This could cause api request failures in the best case scenario.

shreddedbacon:
> Since this replaces pods one at a time, it's possible that some pods are available to accept api requests but don't know about the db schema changes yet. This could cause api request failures in the best case scenario.

This already happens with the current strategy if you don't scale the API to 1 pod before applying the update. There will be X old pods, and X new pods.

I'm open to other methods that can reduce the requirement of having to scale the API to 1 before doing an apply though.

smlx:
Trying to do this by relying on the rollout strategy will break if e.g. a pod restarts/reschedules for another reason (e.g. spot node disappears) during the migration.

IMO it would be more robust to make the `initContainer` exit with a non-zero exit code if it detects an existing lock. That way the other pods would enter a CrashLoop until whichever pod has the lock completes.

shreddedbacon:
> Trying to do this by relying on the rollout strategy will break if e.g. a pod restarts/reschedules for another reason (e.g. spot node disappears) during the migration.
> 
> IMO it would be more robust to make the `initContainer` exit with a non-zero exit code if it detects an existing lock. That way the other pods would enter a CrashLoop until whichever pod has the lock completes.

That works for me too, I just have no idea how to do this with knex. Maybe @tobybellwood or someone more familiar with it can do that

shreddedbacon:
Closing this so someone more familiar with knex locking can resolve it

PR #554, opened 2023-03-15
Update db migration handling and lagoon-core appVersion v2.14.1

### LINKED ISSUE #553: Fix API init container database migration lock issue
When applying an update to Lagoon-core that contains knex db migrations, the API rollout gets held up because init containers start at the same time and they all cause a migration lock and then none of the migrations run.

The only fix for this is to scale the API deployment to 0, clear the migration table locks from the api-db using `DELETE FROM knex_migrations_lock;` and then scaling the API deployment to 1, waiting for the pod to be running, then scaling the API deployment to the previous replica count.

This is not a good experience.

An attempt was made in #552 but this has its own issues as mentioned in comments.

This needs to be fixed ASAP, lots of planned changes to lagoon-core will require DB migrations and this problem will be more of an issue.

CGoodwin90:
Moves the functionality of the initContainer into a Job utilizing a chart hook.
This should the resolve the migration lock issues noted in #553.

closes #553

### DIFF
--- a/charts/lagoon-core/templates/api.migratedb.job.yaml
+++ b/charts/lagoon-core/templates/api.migratedb.job.yaml
@@ -19,7 +19,7 @@ spec:
           {{- toYaml .Values.api.securityContext | nindent 10 }}
       terminationGracePeriodSeconds: 120
       containers:
-      - name: api-init
+      - name: api-migratedb
         args:
         - ./node_modules/.bin/knex migrate:list --cwd /app/services/api/database;
           ./node_modules/.bin/knex migrate:latest --cwd /app/services/api/database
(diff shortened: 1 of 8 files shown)


charts/lagoon-core/templates/api.migratedb.job.yaml
```diff
@@ -0,0 +1,54 @@
...
+    metadata:
+      labels:
+        {{- include "lagoon-core.api.selectorLabels" . | nindent 8 }}
+    spec:
+      restartPolicy: Never
+      securityContext:
+          {{- toYaml .Values.api.securityContext | nindent 10 }}
+      terminationGracePeriodSeconds: 120
+      containers:
+      - name: api-init
```
tobybellwood, line 22:
```suggestion
      - name: api-migratedb
```

CGoodwin90: Missed this! Thanks

PR #558, opened 2023-03-30
Update lagoon-opensearch-sync

smlx:
Fix a typo in the lagoon-opensearch-sync secret template, and bump the version to v0.5.0.


tobybellwood (approved):
nice pickup

PR #561, opened 2023-04-19
Update dockerhost deployment

CGoodwin90:
In conjunction with https://github.com/uselagoon/lagoon-service-images/pull/25

- Removes cronjob env vars as they've been re-written in Go via https://github.com/uselagoon/lagoon-service-images/pull/25
- Specifies protocol and port for use with the Go docker client

### DIFF
--- a/charts/lagoon-remote/templates/docker-host.deployment.yaml
+++ b/charts/lagoon-remote/templates/docker-host.deployment.yaml
@@ -56,7 +56,7 @@ spec:
           value: {{ . | quote }}
         {{- end }}
         - name: DOCKER_HOST
-          value: localhost
+          value: tcp://localhost:2375
         - name: REGISTRY
           value: {{ .Values.dockerHost.registry | quote }}
         - name: REPOSITORY_TO_UPDATE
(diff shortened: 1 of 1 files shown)


charts/lagoon-remote/templates/docker-host.deployment.yaml
```diff
@@ -56,18 +56,23 @@ spec:
...
-            */15 * * * * /lagoon/cronjob.sh "/update-images.sh"
+        - name: PRUNE_SCHEDULE
+          value: {{ .Values.dockerHost.pruneImagesSchedule | quote }}
+        - name: REMOVE_EXITED_SCHEDULE
+          value: {{ .Values.dockerHost.removeExitedSchedule | quote }}
+        - name: UPDATE_IMAGES_SCHEDULE
+          value: {{ .Values.dockerHost.updateImagesSchedule | quote }}
+        - name: DANGLING_FILTER
+          value: {{ .Values.dockerHost.danglingFilter | quote }}
         ports:
```
tobybellwood, line 76: we should add the ability to pass additionalEnvs here

charts/lagoon-remote/templates/docker-host.deployment.yaml
```diff
@@ -56,18 +56,23 @@ spec:
           value: {{ . | quote }}
         {{- end }}
         - name: DOCKER_HOST
-          value: localhost
+          value: tcp://localhost:2375
```
tobybellwood, line 59: we should allow the passing of a dockerHost.host here, but with this as a fallback

PR #569, opened 2023-05-24
Add option to post process router logs

smlx:
This feature allows injecting snippets of fluentd configuration into the
router logs processing pipeline directly before the logs are sent to the
output.

This feature is designed to allow injecting logic to manipulate or add
fields to router log records which make the log records easier to use in
Opensearch Dashboards. For example this can be used to inject a boolean
field into log records which can be used instead of a slow Opensearch
regex search.


shreddedbacon (approved):
Fairly straightforward change, no changes to defaults and clearly outlines what is happening if the values are defined.

Issue #570, opened 2023-05-26
add servicemonitor for fluentbit


domcyrus:
At the moment there is no way to configure a servicemonitor for fluentbit for the lagoon logging component. It does expose prometheus metrics on /api/v1/metrics/prometheus and would make it possible to properly monitor the component.

https://docs.fluentbit.io/manual/administration/monitoring

PR #571, opened 2023-05-29
Expose the metrics configuration of the Logging Operator subchart

smlx:
This change exposes the metrics configuration of the Logging Operator subchart and enables the fluentbit daemonset's ServiceMonitor by default.

Closes: #570


tobybellwood (approved):
yah - this looks good to me - it's flexible enough to support any additional config as needed under the serviceMonitor if required later

tobybellwood (approved):
yah - this looks good to me - it's flexible enough to support any additional config as needed under the serviceMonitor if required later

PR #573, opened 2023-06-02
Bump lagoon-opensearch-sync

smlx:
v0.7.1 fixes a minor bug.


shreddedbacon (approved):
Since this is disabled by default, anyone using it should already be using opensearch.

PR #576, opened 2023-06-06
Updates insights-remote clusterrole to access ns and secrets

bomoko:
Newer versions of insights-remote need to CRUD secrets as well as read namespace details - this is to support a mechanism that injects tokens into environments that can be used to write facts and problems to a local web services.
The token is used to secure the process and is linked to the environment name.


tobybellwood:
the insights-remote chart is deprecated, and is now tucked under lagoon-remote - these changes will need to be made there

bomoko:
Thanks @tobybellwood

PR #578, opened 2023-06-07
Feature/remote insights update

bomoko:
This PR changes the insights remote charts to

1. update the permissions for the clusterrole - we need to list namespaces and CRUD secrets
2. introduces a service that is used to expose a web endpoint to which users can write


charts/lagoon-remote/templates/insights-remote.service.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "lagoon-remote.insightsRemote.fullname" . }}
+  labels: {{- include "lagoon-remote.insightsRemote.labels" . | nindent 4 }}
```
tobybellwood, line 5: I would, out of habit, put anything nindented on the line it's to be added in (ie 4sp in, in the line below)

tobybellwood (approved):
Tested working

PR #579, opened 2023-06-08
Bump remote-controller to version 0.13.0

shreddedbacon:
This bumps the remote-controller to version 0.13.0 which has a number of changes. One is the ability to now inject some additional variables into builds that are prefixed with `LAGOON_CONFIG_`. These are to deprecate the existing `TASK_API/TASK_SSH` prefixed variables.

As such, this chart also needs to be updated to stage the deprecation of those variables. By default if anyone has already defined the `taskAPI/taskSSH` values, these will be consumed. But if a user chooses to do so, they can define the new values of `lagoonAPIHost`, `lagoonTokenHost/Port` and `sshPortalHost/Port` to point to the new services that will eventually replace the legacy SSH service in Lagoon.

### DIFF
--- a/charts/lagoon-build-deploy/Chart.yaml
+++ b/charts/lagoon-build-deploy/Chart.yaml
@@ -16,11 +16,11 @@ kubeVersion: ">= 1.21.0-0"
 
 type: application
 
-version: 0.22.0
+version: 0.22.1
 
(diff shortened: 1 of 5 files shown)


tobybellwood (approved):
LGTM

Issue #589, opened 2023-06-22
Tolerations/Affinity can't be configured on the FluentD deployment in lagoon-logging


FleetAdmiralButter:
We'd like to configure the FluentD deployment in the lagoon-logging chart to deploy on a specific node group.

The logging.banzaicloud.io template doesn't currently support this.

PR #590, opened 2023-06-22
Ability to configure toleration/affinity on FluentD deployment

FleetAdmiralButter:
Closes: #589 

Extends the logging.banzaicloud.io template and values.yaml to allow for configuration of tolerations/affinity on the FluentD deployment.


charts/lagoon-logging/Chart.yaml
```diff
@@ -19,7 +19,7 @@ type: application
 # time you make changes to the chart and its templates, including the app
 # version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.78.0
+version: 0.79.0
```
tobybellwood, line 22: you'll also need to replace the `artifacthub.io/changes` section to explain what this change is for

charts/lagoon-logging/values.yaml
```diff
@@ -251,6 +251,11 @@ enableDefaultForwarding: true
 # Set how many fluentd pods should be running
 fluentdReplicaCount: 3
 
+fluentd:
```
tobybellwood, line 254: It's a pain, but you should probably use the unnested concatenated names here to avoid confusion - eg fluentdTolerations and fluentdAffinity, as per the fluentdReplicaCount and fluentdMetrics

tobybellwood (commented):
thanks! Couple of minor nits, but looks ok - will test a bit further

FleetAdmiralButter:
Thanks @tobybellwood, I've updated the pull request

PR #596, opened 2023-07-26
Monitor Lagoon Broker with more details

Schnitzel:
This tells Prometheus to scrape the lagoon broker more detailed, which allows to monitor the amount of messages and consumers per queue


tobybellwood:
closing - merging into #598

PR #601, opened 2023-08-09
convert api-redis to redis-persistent

tobybellwood:
This PR adds a persistence.enabled option to api-redis that, when enabled:
* creates the PVC, volume  and mount
* sets the necessary variable to enable the persistence settings

I've set the default storage size to 1Gi and disabled it in CI.

### DIFF
--- a/charts/lagoon-core/templates/api-redis.deployment.yaml
+++ b/charts/lagoon-core/templates/api-redis.deployment.yaml
@@ -36,6 +36,10 @@ spec:
             secretKeyRef:
               name: {{ include "lagoon-core.api.fullname" . }}
               key: REDIS_PASSWORD
+        {{- if .Values.apiRedis.persistence.enabled }}
+        - name: FLAVOR
+          value: persistent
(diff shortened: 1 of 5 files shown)


charts/lagoon-core/templates/api-redis.deployment.yaml
```diff
@@ -63,3 +78,14 @@ spec:
...
+  volumeClaimTemplates:
+  - metadata:
+      name: {{ include "lagoon-core.apiRedis.fullname" . }}-data
+    spec:
+      accessModes:
+      - ReadWriteOnce
+      resources:
+        requests:
+          storage: {{ .Values.apiRedis.storageSize | quote }}
+  {{- end }}
```
shreddedbacon, line 91: I don't think this is required

tobybellwood: Ah yes, could be a statefulSet hangover

tobybellwood:
merged in #603

PR #602, opened 2023-08-11
Upgrade nats subchart to v1.2

smlx:
This PR upgrades the NATS subchart dependency to 1.0.

This is a __breaking change__. The NATS chart has completely restructured its configuration between v0.x and v1.x.
So by necessity the NATS subchart configuration of `lagoon-core` and `lagoon-remote` also had to be restructured.
In general, the major upstream change affecting the Lagoon charts is that `nats.*` values have mostly been moved to `nats.config.*`.

In particular, the following values required by the Lagoon charts have changed:

Lagoon Core:

| Old path                                 | New path                              |
| ---                                      | ---                                   |
| `nats.cluster.name`                      | `nats.config.cluster.name`            |
| `natsConfig.tls.secretData."server.crt"` | `natsConfig.tls.secretData."tls.crt"` |
| `natsConfig.tls.secretData."server.key"` | `natsConfig.tls.secretData."tls.key"` |

Lagoon Remote:

| Old path                                 | New path                              |
| ---                                      | ---                                   |
| `nats.cluster.name`                      | `nats.config.cluster.name`            |
| `natsConfig.tls.secretData."client.crt"` | `natsConfig.tls.secretData."tls.crt"` |
| `natsConfig.tls.secretData."client.key"` | `natsConfig.tls.secretData."tls.key"` |

In addition, any other NATS subchart configuration options that may have been added (anything under `nats.*` in the `values.yaml` of `lagoon-core` and `lagoon-remote`) will most likely require updating.

There are upstream notes on the changes in the NATS chart [here](https://github.com/nats-io/k8s/blob/7fcc48ef44a4ecc686f15becec9cc5e283853bbd/helm/charts/nats/UPGRADING.md#update-nats-config-to-new-valuesyaml-schema).

The other change in this PR is that the client certificates presented by Lagoon Remote are now validated by NATS in Lagoon Core (they were not previously validated correctly).
This should not be a breaking change if the previous instructions for generating certificates were followed correctly.

Also note that due to the breaking change this is a major version bump for `lagoon-core` from `v1.x` to `v2.x`.
This is not a major version bump for `lagoon-remote` because that is still at `v0.x`.

### DIFF
--- a/charts/lagoon-core/Chart.lock
+++ b/charts/lagoon-core/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: nats
   repository: https://nats-io.github.io/k8s/helm/charts/
-  version: 0.19.17
-digest: sha256:9c58fc4ddeec7b86f5ef2cf1996a48a7e09d9bd4aa149971e2525a6f05649bf8
-generated: "2023-07-28T09:49:46.220986689+08:00"
+  version: 1.0.3
+digest: sha256:437a0a6fea805b16ec2d292a0cf9bbba8084e09448c9d9b5ae7fcae1be14bb47
+generated: "2023-09-20T12:18:17.665626036+08:00"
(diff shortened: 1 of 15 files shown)


tobybellwood (approved):
Tested working, will add notes to release to cover upgrade process.

PR #605, opened 2023-08-21
Schedule Logging Pods also on infra nodes

### LINKED ISSUE #604: lagoon-logging fluentbit are not scheduled on infra nodes
Lagoon installations have now more and more `lagoon.sh/infra` tainted nodes, we should deploy the lagoon-logging fluentbit pods on them by default

Schnitzel:
closes #604 604


tobybellwood:
working in #620

PR #607, opened 2023-09-07
Always create sshPortal authorization resources and add a service account token resource

rocketeerbkw:
In order to keep the legacy SSH service in lagoon-core working in recent k8s clusters, a non-expiring token is needed. A sufficiently scoped role and binding already exists for the ssh-portal, so this PR adds the service account token secret that can be used in lagoon-core. Since this is needed even if ssh-portal isn't used, the auth related resources are changed to always be created, even if the ssh-portal service isn't enabled.

### DIFF
--- a/charts/lagoon-remote/templates/ssh-portal.clusterrole.yaml
+++ b/charts/lagoon-remote/templates/ssh-portal.clusterrole.yaml
@@ -1,4 +1,3 @@
-{{- if .Values.sshPortal.enabled -}}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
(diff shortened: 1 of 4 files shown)


smlx:
The documentation for the k8s API used in this PR is here: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#manually-create-an-api-token-for-a-serviceaccount

Could we put the creation of this secret behind a flag so that people have to understand the step they are taking with this? e.g. maybe a top-level `createLegacySSHServiceAccountToken: false` in the values file with an explanatory comment?

rocketeerbkw:
Superseded by https://github.com/uselagoon/lagoon-charts/pull/608

PR #610, opened 2023-09-19
fix: upgrade mariadb chart to fix upgrade bug

smlx:
This change allows `make install-lagoon-remote` to be run more than once.

Previously we would hit https://github.com/bitnami/charts/issues/15093.

The specific chart version that this PR upgrades to is the latest version of the chart which still uses MariaDB 10.11.x (the LTS minor version of v10). Later versions of the chart upgrade to MariaDB 11, which is outside the scope of this bugfix.


tobybellwood (approved):
nice, thx

PR #611, opened 2023-09-19
Update README with local caching proxy

smlx:
This PR adds documentation of a nice hack that means all containers of `lagoon-core` will start running in <60s when running locally in `kind` (once the cache is warm). This helps to make local dev of a full lagoon installation a much nicer experience.

screenshot_2023-09-19-153724


smlx:
Once https://github.com/uselagoon/lagoon/pull/3499 lands, combined with this, we should be able to go from zero to lagoon-core much faster :rocket:

PR #620, opened 2023-10-17
Logging updates

### LINKED ISSUE #604: lagoon-logging fluentbit are not scheduled on infra nodes

tobybellwood:
This PR collects a number of improvements to the logging setup in Lagoon.

it:
* updates the versions of the images used in the logs-dispatcher and logs-concentrator
* ensures that Loggin pods also schedule on infra nodes
* Allows users to configure toleration/affinity for FluentD deployments

closes #604 #589

### DIFF
--- a/charts/lagoon-logging/values.yaml
+++ b/charts/lagoon-logging/values.yaml
@@ -251,6 +251,10 @@ enableDefaultForwarding: true
 # Set how many fluentd pods should be running
 fluentdReplicaCount: 3
 
+fluentdTolerations: []
+
+fluentdAffinity: {}
+
 # Add tolerations to the fluentbit daemonset so that it runs on e.g.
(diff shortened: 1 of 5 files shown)


charts/lagoon-logging/values.yaml
```diff
@@ -251,6 +251,10 @@ enableDefaultForwarding: true
 # Set how many fluentd pods should be running
 fluentdReplicaCount: 3
 
+fluentdTolerations: []
```
smlx, line 254: Could we add a comment on these new fields describing what they do and why you might want to use them? e.g. similar to the `fluentbitTolerations` comment.

PR #621, opened 2023-10-25
Feature/insights handler trivy server

bomoko:
This PR introduces a Trivy server component for problems generation for lagoon-core.

If enabled (via setting `.Values.insightsTrivy.enabled=true`) this will change the insights-handler's deployment to include  env vars to enable Trivy scans of incoming SBOMS, as well as spinning up a Trivy server instance.


tobybellwood (approved):
This has been tested working.

Nesting the trivy under the handler makes sense, as there isn't need for it without the handler.

PR #630, opened 2023-11-14
chore: remove dioscuri subchart from lagoon-remote

shreddedbacon:
Note: leaving this as a draft for now until removal of dioscuri is confirmed.
The active-standby-kubernetes test will fail until the updated task and API changes are merged in Lagoon upstream.

Removing the dioscuri subchart for lagoon-remote as it will be deprecated soon.


tobybellwood (approved):
merging to 218 prerelease now the task is updated to resolve the conflict

PR #631, opened 2023-11-14
add variable to control ssh portal and nats

tobybellwood:


shreddedbacon:
Is this still required? I don't think it is?

PR #637, opened 2023-12-07
chore: add clusterroles for tasks and builds to admin and edit

shreddedbacon:
Just adding the missing clusterroles to the admin role to allow the lagoon namespace deployer service account to interact with builds and tasks within the namespace as required

Currently the deployer service account can interact with all other resources within the namespace during a build, and being able to interact with other builds or tasks that may exist in the namespace would be useful for being able to handle previous build or task cleanups within a build process.

### DIFF
--- a/charts/lagoon-build-deploy/Chart.yaml
+++ b/charts/lagoon-build-deploy/Chart.yaml
@@ -16,11 +16,11 @@ kubeVersion: ">= 1.23.0-0"
 
 type: application
 
-version: 0.26.2
+version: 0.26.3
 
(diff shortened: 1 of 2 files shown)


shreddedbacon:
It would be nice to get this into a lagoon-remote update as part of lagoon-remote 0.87.0 in the 2.17 branch

tobybellwood (approved):
This is a neat change, allowing correctly permissioned namespace-scoped serviceaccounts the correct permissions to access all builds and task crds in their namespace

PR #638, opened 2023-12-14
fix: avoid installing the registry multiple times

smlx:
While the fill-test-ci-values target's dependency on install-registry was correctly manipulated by the SKIP_INSTALL_REGISTRY variable, the install-lagoon-build-deploy target had a hard-coded dependency on install-registry. The result was that the registry would get installed multiple times anyway via the fill-test-ci-values target via an indirect dependency.

Fix that by removing the dependency that install-lagoon-build-deploy has on install-registry when SKIP_INSTALL_REGISTRY=true.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -261,7 +261,7 @@ install-lagoon-remote: install-lagoon-build-deploy install-lagoon-core install-m
 # Do not install without lagoon-core
 #
 .PHONY: install-lagoon-build-deploy
-install-lagoon-build-deploy: install-lagoon-core install-registry
+install-lagoon-build-deploy: install-lagoon-core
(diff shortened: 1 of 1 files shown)


shreddedbacon:
running `make TESTS=[nginx] k3d/test` in `uselagoon/lagoon`. Not hugely noticeable difference. I think it is probably worth merging though

before: 
```
real	21m31.993s
user	0m46.142s
sys	0m8.614s
```
with this branch:
```
real	20m11.619s
user	0m44.983s
sys	0m8.093s
```

Only shaved off a small amount of time

tobybellwood (approved):
straightforward enough, merging to prerelease

PR #641, opened 2024-02-15
Support Keycloak migrations via the API init-container

shreddedbacon:
This PR adds the keycloak migrations to the initcontainer run in the API

https://github.com/uselagoon/lagoon/pull/3612


tobybellwood (approved):
Straightforward enough addition

PR #642, opened 2024-02-15
CDN service annotation support

smlx:
Newer versions of the AWS load balancer controller [change the default behaviour](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/service/annotations/#lb-scheme) of `LoadBalancer` objects to be internal (private IP address) rather than internet-facing. The interface for making a `LoadBalancer` internet-facing is now an annotation on the k8s service object.

So this PR adds support for annotations on the CDN service, which is the only one intended to be used as a `LoadBalancer`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,18 @@ Documentation on probes for pod startup is [here](https://kubernetes.io/docs/con
 
 ### Run chart-testing (lint) locally
 
+```bash
+docker run --rm --interactive --detach --network host --name ct "--volume=$(pwd):/workdir" "--workdir=/workdir" --volume=$(pwd)/default.ct.yaml:/etc/ct/ct.yaml quay.io/helmpack/chart-testing:latest cat
+docker exec ct git config --global --add safe.directory /workdir
+docker exec ct ct lint
 ```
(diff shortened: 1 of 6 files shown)


tobybellwood:
I stumbled into issues testing locally, as the kubectl container couldn't build (the alpine pkgs wouldn't install kubectl and jq) - but I replaced it with `uselagoon/tests:latest` - which has them preinstalled, and it worked fine.

```
------------------------------------------------------------------------------------------------------------------------
==> Logs of container lagoon-logging-7l8a9hjbxf-logs-dispatcher-test-cdn-service-annotations
------------------------------------------------------------------------------------------------------------------------
fetch https://dl-cdn.alpinelinux.org/alpine/v3.19/main/x86_64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.19/community/x86_64/APKINDEX.tar.gz
WARNING: fetching https://dl-cdn.alpinelinux.org/alpine/v3.19/main: could not connect to server (check repositories file)
WARNING: fetching https://dl-cdn.alpinelinux.org/alpine/v3.19/community: could not connect to server (check repositories file)
ERROR: unable to select packages:
  jq (no such package):
    required by: world[jq]
  kubectl (no such package):
    required by: world[kubectl]
    ```

tobybellwood (approved):
Works as designed, tests will need a bit more going through before we can adopt them, but ran locally ok, and not enough to block release.

PR #645, opened 2024-02-23
chore: add legacy registry passthrough support

shreddedbacon:
Just adds the legacy registry passthrough support to the build-deploy chart since it is going to be removed from core https://github.com/uselagoon/lagoon/pull/3659


tobybellwood (approved):
I'm happy with this approach - it makes more sense here than in core.

PR #653, opened 2024-04-11
update Lagoon appVersion to v2.18.2

tobybellwood:
This PR updates Lagoon to appVersion v2.18.2
It also updates storage-calculator to v0.5.3
It also adds the logic to API.deployment to present the ssh endpoints for https://github.com/uselagoon/lagoon/pull/3632

### DIFF
--- a/charts/lagoon-core/templates/api.deployment.yaml
+++ b/charts/lagoon-core/templates/api.deployment.yaml
@@ -175,6 +175,16 @@ spec:
         - name: S3_FILES_REGION
           value: {{ . | quote }}
         {{- end }}
+      {{- if .Values.sshTokenEndpoint }}
+        - name: SSH_TOKEN_ENDPOINT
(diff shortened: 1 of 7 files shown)


charts/lagoon-core/templates/api.deployment.yaml
```diff
@@ -68,6 +68,14 @@ spec:
           {{- end }}
         - name: REDIS_HOST
           value: {{ include "lagoon-core.apiRedis.fullname" . }}
+        - name: SSH_TOKEN_ENDPOINT
+          value: {{ .Values.sshTokenEndpoint | quote }}
```
shreddedbacon, line 72: i think this will fail unless it is defined. but the default value is commented out?

tobybellwood: yuup - wrapped in an "if" now

shreddedbacon (approved):
Looks right to me

PR #654, opened 2024-04-18
chore: Use dedicated service for metrics

anothertobi:
When the main service is used as a load balancer service to allow public
access through mTLS, the metrics should not be exposed on the same
service.

### DIFF
--- a/charts/lagoon-logs-concentrator/Chart.yaml
+++ b/charts/lagoon-logs-concentrator/Chart.yaml
@@ -19,7 +19,7 @@ type: application
 # time you make changes to the chart and its templates, including the app
 # version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.47.0
+version: 0.48.0
 
 # This section is used to collect a changelog for artifacthub.io
 # It should be started afresh for each release
@@ -27,4 +27,4 @@ version: 0.47.0
 annotations:
   artifacthub.io/changes: |
     - kind: changed
-      description: update uselagoon/logs-concentrator image to v3.2.0
+      description: use dedicated service for metrics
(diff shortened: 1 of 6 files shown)


charts/lagoon-logs-concentrator/templates/metrics-service.yaml
```diff
@@ -0,0 +1,22 @@
...
+apiVersion: v1
+kind: Service
+metadata:
+{{- with .Values.metrics.service.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+{{- end }}
+  name: {{ include "lagoon-logs-concentrator.fullname" . }}-metrics
+  labels:
+    service: metrics
```
smlx, line 11: As discussed, maybe this should be updated to `metrics-only=true` to match the other charts? But I'm not too fussed. Just whatever is most consistent with other servicemonitor selectors.

charts/lagoon-logs-concentrator/Chart.yaml
```diff
@@ -19,7 +19,7 @@ type: application
 # time you make changes to the chart and its templates, including the app
 # version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.47.0
+version: 0.48.0
 
 # This section is used to collect a changelog for artifacthub.io
```
smlx, line 24: The CI failure is due to this field not being updated. This field must be updated on every PR because it generates the changelog on artifacthub. The format is documented [here](https://artifacthub.io/docs/topics/annotations/helm/#supported-annotations).

smlx (changes requested):

charts/lagoon-logs-concentrator/templates/metrics-service.yaml
anothertobi: Updated it so that it aligns with the other charts in this repository.

smlx (approved):
LGTM thanks!

PR #664, opened 2024-05-28
build: update remote-controller image to v0.15.6

tobybellwood:


charts/lagoon-build-deploy/Chart.yaml
```diff
@@ -16,11 +16,11 @@ kubeVersion: ">= 1.23.0-0"
 
 type: application
 
-version: 0.26.4
```
shreddedbacon, line 19: Could this have been a patch version?

shreddedbacon (approved):
All good here, just a question on the version number, but I'm not fussed if it is ignored :smile:

charts/lagoon-build-deploy/Chart.yaml
tobybellwood: possibly - but I always tend towards a minor release on an appversion bump, and reserve patches for minor functionality or fixups between?

PR #666, opened 2024-05-30
feat: added broker-flag-enable pre-update job

tobybellwood:
Adds a pre-update job to ensure that all RabbitMQ feature flags are set on an upgrade as required in https://github.com/uselagoon/lagoon/pull/3670


tobybellwood:
Holding this off until post-lagoon 2.20 so the functionality is in the previous release that we rely on

tobybellwood:
merging in #684

PR #669, opened 2024-06-16
Lagoon v2.20.0 release, Kubernetes 1.25 minimum

tobybellwood:
This PR is a pre-release branch for Lagoon v.2.20.0

It also incorporates a number of other improvements:
* automated `next-release` label testing to remove the need to. manually set testlagoon&main for images
* updated Kubernetes version to 1.30 and associated packages
* Sets the minimum supported version of Kubernetes to 1.25 (as per some of the upstream charts for other components)
* Streamlines the install process for the lint/test routines to use built-in binaries where possible.
* It incorporates the local-development changes from https://github.com/uselagoon/lagoon/pull/3710


shreddedbacon (approved):
From what I understand of the changes, this seems like it is mostly testing changes rather than impact changes to lagoon core v2.20.0 specifically.

Of the changes that relate to lagoon core v2.20.0, being the api sidecar and other version bumps this seems fine.

PR #671, opened 2024-06-20
chore: update crd for builds and tasks

shreddedbacon:
Updates crds to include the organization fields in the spec


tobybellwood (approved):
LGTM

PR #676, opened 2024-07-30
fix: lagoon-remote-ssh-core scale permissions

rocketeerbkw:
In [Lagoon 2.20.0](https://github.com/uselagoon/lagoon/releases/tag/v2.20.0), the `kubectl` binary was upgraded, and it appears to use a different method of scaling deployments that requires an extra permission. This PR adds the ability for the lagoon-remote ssh-core service account to `patch`, in addition to `get` and `update`.


shreddedbacon (approved):
Thanks for digging this one up, solves the problem!

PR #677, opened 2024-07-30
Feature - Enable post-build scanning in insights-remote

bomoko:
This PR supports the introduction of Post-build image scans in the insights-remote controller.

It supports the changes introduced in https://github.com/uselagoon/insights-remote/pull/43

Essentially, the changes are around the cluster role to expand permissions (we now need to be able to CRUD `pods`), as well as allowing customization of the scanner image that's used in the scanning process.

* The post-build image scanning is disabled by default, this is changed with `insightsRemote.enableBuildScanning` set to `true`
* overriding the scan image can be done by explicitly setting `insightsRemote.buildScannerImage` to some image

### DIFF
--- a/charts/lagoon-remote/templates/docker-host.networkpolicy.yaml
+++ b/charts/lagoon-remote/templates/docker-host.networkpolicy.yaml
@@ -16,6 +16,14 @@ spec:
         matchExpressions:
         - key: lagoon.sh/buildName
           operator: Exists
+    - namespaceSelector:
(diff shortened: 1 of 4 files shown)


bomoko:
> Does this require a version bump of `insights-remote` image too? Or are the changes in this compatible with current and next versions?

The changes were compatible, but now we have the version - so it makes sense to bump.

shreddedbacon (approved):

PR #678, opened 2024-08-07
feat: support for static hostkeys in ssh core

### LINKED ISSUE #655: core ssh service host keys
It looks like we could use secret mounts to inject host keys into the core ssh service to prevent host key warnings

https://github.com/uselagoon/lagoon/blob/main/services/ssh/etc/ssh/sshd_config#L3-L5

shreddedbacon:
SSH portal allows for keys to be defined for host key verification. Lagoon-core SSH service does not and randomly generates a new key each time, this provides a bad experience and forces users to ignore the host key checks.

This change allows for keys to be defined on the core service now.

The core ssh service defines has the following https://github.com/uselagoon/lagoon/blob/2b4283b707a74b8d677c6f32d595fbcf51e613fb/services/ssh/etc/ssh/sshd_config#L3-L5

If they're not provided by the chart, then the core service will generate them each time a pod restarts (not ideal)

closes #655


shreddedbacon:
closed as part of #679

PR #679, opened 2024-08-08
update Lagoon appVersion to v2.20.1

tobybellwood:
This PR updates the lagoon-core chart for Lagoon release https://github.com/uselagoon/lagoon/releases/tag/v2.20.1

It also adds the ability to set hostkeys on the SSH service in core, to avoid the possibility of rotating keys causing known_hosts issues

To generate the keypairs, run the following commands, and look at the `ssh` section in https://github.com/uselagoon/lagoon-charts/blob/main/charts/lagoon-core/ci/linter-values.yaml for formatting
```
	ssh-keygen -q -t ed25519 -N '' -C '' -f ed25519
	ssh-keygen -q -t ecdsa -N '' -C '' -f ecdsa
	ssh-keygen -q -t rsa -N '' -C '' -f rsa
```


charts/lagoon-core/ci/linter-values.yaml
```diff
@@ -163,6 +163,67 @@ ssh:
   resources:
     requests:
       cpu: "10m"
+  hostKeys:
+    rsaPub: "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQDTU9JXClTTQQB2QvsNegwWOwxdZNyswjDrWO94c7Gej+WkeFC6iYgIkALTFXP+/uSgMdcxISnAqaFial7j2owmnGKbH9MTTmSS68I8FZRN/KAsDTVqzOR8bcwo36uS6cjcvQJevFpKJk/SWDMozqRgWK2I3QfkkCvGyRA9x/kq9BKFORjMzcbRnSTv3VPsXJgJSIQGYp+76TTNIT28h+WezGVx1Yf+sKmhJjgVnMcvim66KW+70fgeaScf5sgsWBX+31T5RkVqarhW+FBcHuFk5N3bmsG7k+894nssqDvVPXEKly386q4Z9OykaYv+q/o91wAA+UwUg7Pb+6D7X80cjxLBENoH2HSraKTj+Z0e6i7usCu2Zd/xUGUiSmLtfQ1hP83HeeBEWUeVijs70/gtV9Xj8GXncaKoKyV/Ws6w2nCK4b1UgfV0bMLDdgjeA0+KFA8RSckoVWYtFh3zUxG/vGeAVZkgncKShEJMgeQZJ9DzutoO2ZZsYTth1C8rab0="
```
smlx, line 167: I don't think you need the public key. Just the private is enough for `HostKey`.

tobybellwood: If you don't have a matching public key, the error log spits out a complaint every single time!

shreddedbacon (approved):
Yep, pretty simple changes overall. 

The concern around the SSH host key/public key injection is valid, but it will need to be revisited in the future when we look to handling other issues in the core `ssh` service in `uselagoon/lagoon` to improve as time permits.

PR #681, opened 2024-09-02
Update lagoon-logging chart, logs-concentrator and logs-dispatcher images

tobybellwood:
Currently, this PR updates the charts to 4.9.1 which includes https://github.com/kube-logging/logging-operator/pull/1794

### DIFF
--- a/charts/lagoon-logging/Chart.lock
+++ b/charts/lagoon-logging/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: logging-operator
   repository: oci://ghcr.io/kube-logging/helm-charts
-  version: 4.9.0
-digest: sha256:6c06a155e62a3716a1d549187a29e4fa8cdf59ddcf3bddec58e2abcb07ffa27d
-generated: "2024-08-15T16:08:40.297734358+10:00"
+  version: 4.9.1
+digest: sha256:26b05aafbf9e1d92adfd10664ac225d8f07adc00e497369d563c4b12d445c108
+generated: "2024-09-02T11:31:43.731650736+10:00"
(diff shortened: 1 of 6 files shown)


smlx:
Looks like it might be [this field](https://kube-logging.dev/docs/configuration/crds/v1beta1/fluentbit_types/#filterkubernetes-namespace_labels) in the fluentbit CRD which changed to on by default in the latest version of the operator in [this MR](https://github.com/kube-logging/logging-operator/pull/1794/files).

So if we explicitly set this to `off` it might revert to previous behaviour?

tobybellwood:
> Looks like it might be [this field](https://kube-logging.dev/docs/configuration/crds/v1beta1/fluentbit_types/#filterkubernetes-namespace_labels) in the fluentbit CRD which changed to on by default in the latest version of the operator in [this MR](https://github.com/kube-logging/logging-operator/pull/1794/files).
> 
> So if we explicitly set this to `off` it might revert to previous behaviour?

yup - I'm thinking we'll have to bundle a modified CRD in there - same as clusterFlow

tobybellwood:
ah, we'll have to do that anyway to move the fluentbit config out of the logging CRD
```
{"level":"info","ts":"2024-09-02T08:22:13Z","logger":"logging","msg":"WARNING fluentbit definition inside the Logging resource is deprecated and will be removed in the next major release","logging":"lagoon-logging"}

and

{"level":"info","ts":"2024-09-02T08:25:09Z","logger":"logging.fluentbit-legacy","msg":"Notice: fluentbit `network` settings have been configured automatically to adapt to multiple aggregator replicas. Configure it manually to avoid this notice.","logging":"lagoon-logging"}
```

PR #682, opened 2024-09-08
change: set keycloak_frontend_url for external keycloak

shreddedbacon:
This adds the same `KEYCLOAK_FRONTEND_URL` variable that is used by keycloak itself to set the frontend, to inject into the API to be used in the `.well-known` discovery endpoint.

This adds the `/auth` path to this URL so that tools will be able to auto detect the correct path that keycloak is configured with correctly.

If in the future this `/auth` path is removed from being [hardcoded](https://github.com/uselagoon/lagoon/blob/main/services/keycloak/entrypoints/default-keycloak-entrypoint.sh#L10) as a path, tools will be able to rediscover keycloak without the relative path to `/auth`


shreddedbacon:
Probably best to merge into https://github.com/uselagoon/lagoon-charts/pull/684

PR #683, opened 2024-09-09
feat: add support for setting keycloak redirect URIs for lagoon UI

rocketeerbkw:
Support was added to set valid redirect URIs for the lagoon UI client in https://github.com/uselagoon/lagoon/pull/3640. This PR adds this setting to the charts. It defaults to the UI ingress, but can be overridden by `keycloakLagoonUIClientRedirectUris`.


shreddedbacon:
This is sort of a breaking change. I'm fine with it, but it ***has*** to be called out in the release notes

rocketeerbkw:
I'm not sure this is the best option. If there are more client settings that we have to accomadate later, maybe it's better to have no default and do something like:

```
keycloak:
  clientSettings:
    lagoonUI:
      redirectUris: 
```

shreddedbacon:
> I'm not sure this is the best option. If there are more client settings that we have to accommodate later, maybe it's better to have no default and do something like:
> 
> ```
> keycloak:
>   clientSettings:
>     lagoonUI:
>       redirectUris: 
> ```

There is prior art for this sort of structure too https://github.com/uselagoon/lagoon-charts/blob/main/charts/lagoon-core/values.yaml#L235-L243

We'd obviously need to only handle clients that exist already or are created by Lagoon so that `lagoonUi` can be transformed to `lagoon-ui` in keycloak though.

PR #684, opened 2024-09-09
Lagoon v221

tobybellwood:
This is a multi-chart release:

It is primarily for the release of Lagoon Core v2.21.0, but also includes the following updates:
* Updates Lagoon Insights remote&handler
* Updates SSH Portal and Token service
* Adds a pre-upgrade job to ensure broker feature flags are enabled
* Fixes up Keycloak URLs to take internal traffic
* Updates the versions of Kubernetes that are tested


charts/lagoon-core/templates/api.deployment.yaml
```diff
@@ -153,14 +155,16 @@ spec:
             secretKeyRef:
               name: {{ include "lagoon-core.keycloak.fullname" . }}
               key: KEYCLOAK_API_CLIENT_SECRET
-        - name: KEYCLOAK_URL
+        - name: KEYCLOAK_FRONTEND_URL
           {{- if .Values.keycloakFrontEndURL }}
```
shreddedbacon, line 159: ahh crap, these will need the `/auth` added to them like above

shreddedbacon: https://github.com/uselagoon/lagoon-charts/commit/c84b5b3ae84911e48231aa121aac7cc7fdb0e3e5

i forgot that i was only adding to the init container, not the api :facepalm: bloody templates

charts/lagoon-core/ci/linter-values.yaml
```diff
@@ -142,9 +142,9 @@ backupHandler:
       cpu: "10m"
 
 insightsHandler:
+  enabled: true
```
shreddedbacon, line 145: are we enabling by default now?

charts/lagoon-docker-host/ci/linter-values.yaml
shreddedbacon, line 1: we don't use this docker-host chart in testing lagoon?

tobybellwood: We don't but the kind update enabled the network policy - so this change had to modify it to allow testing

shreddedbacon: easy

shreddedbacon (approved):
Nothing standing out to me here that is concerning. Mostly updates to testing and version bumps.

PR #687, opened 2024-10-01
feat: use statefulset dockerhost

shreddedbacon:
This changes the docker-host from a deployment to a statefulset.

This change also brings individual services per pod. These services will eventually be used by the `remote-controller` to distribute builds across the number of available docker-hosts, this will be used to help with queuing builds better. This statefulset should never be scaled manually, as the service per pod will only be created by helm.

The service has been updated to include [session affinity](https://kubernetes.io/docs/reference/networking/virtual-ips/#session-affinity) capability as a fall back to prevent issues in the event the docker-host replicas are scaled above 1 before the remote-controller changes roll out. This would allow for builds to sort of "round-robbin" as they're started, and that build using session affinity would always be on the docker-host it started on, but the distribution of builds may not be equal across the docker-hosts. The default replica count is still 1.

This is a drop-in replacement for the docker-host, no other changes are required to chart values to make it work. The downside is that the existing docker-host volume will be removed when this is applied, resulting in a completely fresh docker-host volume. This would result in new builds being a bit slower initially while the cache is built up.

Also updated the standalone docker-host chart to statefulset


tobybellwood (approved):
This looks good - tested working and smooth upgrade

PR #689, opened 2024-10-01
feat: change standalone dockerhost to statefulset

shreddedbacon:
Similar to #687 but for the standalone docker-host chart.


shreddedbacon:
added to #687

PR #690, opened 2024-10-03
chore: remove unused banzaicloud-stable repo

shreddedbacon:
We don't appear to use `banzaicloud-stable` anymore for anything, and they are currently experiencing DNS issues that are causing tests to fail. The crds referenced in `lagoon-logging` are installed by kube-logging operator now.

The README in the `lagoon-logging` chart probably needs updating. This just removes banzaicloud-stable from the actions to allow tests to pass.


tobybellwood (approved):
yup - sm0rt

PR #695, opened 2024-10-21
refactor: install isolated lagoon for upgradeability

shreddedbacon:
Supporting changes for https://github.com/uselagoon/lagoon/pull/3829

### DIFF
--- a/.github/workflows/test-suite.yaml
+++ b/.github/workflows/test-suite.yaml
@@ -133,7 +133,9 @@ jobs:
       if: |
         (steps.list-changed.outputs.changed == 'true' && !contains(github.event.pull_request.labels.*.name, 'next-release')) ||
         (contains(github.event.pull_request.labels.*.name, 'needs-testing'))
-      run: make -j8 -O fill-test-ci-values TESTS=[${{ matrix.test }}]
+      run: |
+        make install-lagoon
+        make -j8 -O fill-test-ci-values TESTS=[${{ matrix.test }}]
 
     - name: Helm-install the test fixtures and fill lagoon-test/ci/linter-values.yaml (next-release)
       if: |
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -169,6 +177,7 @@ install-mailpit: install-ingress
 
 .PHONY: install-mariadb
 install-mariadb:
+ifeq ($(INSTALL_MARIADB_PROVIDER),true)
```
smlx, line 180: Makefile conditionals are handled by the preprocessor, before parsing. So this logic effectively means that if `INSTALL_MARIADB_PROVIDER` is not `true`, `install-mariadb` is an empty target. This is unusual for a `Makefile` and a bit confusing because you are still going to be executing the target, it's just going to be empty.

I think the more `make` way to do this would be to make the _dependency_ on the `install-mariadb` conditional? Same goes for the other `install-*` targets.

.github/workflows/test-suite.yaml
```diff
@@ -133,14 +133,19 @@ jobs:
       if: |
         (steps.list-changed.outputs.changed == 'true' && !contains(github.event.pull_request.labels.*.name, 'next-release')) ||
         (contains(github.event.pull_request.labels.*.name, 'needs-testing'))
-      run: make -j8 -O fill-test-ci-values TESTS=[${{ matrix.test }}]
+      run: |
+        make install-lagoon-dependencies
+        make install-lagoon
```
smlx, line 138: Rather than requiring two runs, could these targets be consolidated into `make install-lagoon INSTALL_LAGOON_DEPENDENCIES=true`? I thought that might make it clearer that you can toggle the behaviour rather than knowing that you need to run these commands in order.

Makefile
shreddedbacon: Example?

smlx: Redefine the unconditional `install-lagoon-dependencies` target to only list the unconditional dependencies.

```make
install-lagoon-dependencies: install-registry install-minio install-bulk-storageclass
```

And then conditionally add `install-mariadb` etc.

```make
ifeq ($(INSTALL_MARIADB_PROVIDER),true)
install-lagoon-dependencies: install-mariadb
endif
```

shreddedbacon: NVM

tobybellwood (approved):
Seems to be working, happy to proceed!

PR #700, opened 2024-11-11
feat: add sysctl configure daemonset

shreddedbacon:
To remove https://github.com/uselagoon/build-deploy-tool/blob/d2508efa74871cabe4c477e44bbe87e339d99f5d/internal/servicetypes/opensearch.go#L75-L96 from the opensearch and elasticsearch templates so that no containers require the `privileged: true` value. There needs to be some option to configure any sysctl values as required in `lagoon-remote`.

This adds a `DaemonSet` that allows for setting this value, and potentially other values. This is disabled by default.

Ideally the underlying instances that run kubernetes should have the values updated to remove the need for this, but in cases where it isn't possible to do so, this `DaemonSet` provides the capability.

Linked to https://github.com/uselagoon/build-deploy-tool/pull/395

### DIFF
--- a/charts/lagoon-remote/Chart.yaml
+++ b/charts/lagoon-remote/Chart.yaml
@@ -42,3 +42,5 @@ annotations:
   artifacthub.io/changes: |
     - kind: changed
       description: update dbaas-operator chart dependency to 0.3.1
+    - kind: added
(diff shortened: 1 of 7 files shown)


tobybellwood (approved):
Tested working

PR #705, opened 2024-12-02
chore: add support for keycloak admin api client values

shreddedbacon:
Support for https://github.com/uselagoon/lagoon/pull/3841


tobybellwood (approved):
Straightforward change, just adds a new secret

PR #709, opened 2024-12-12
feat: support for retention policies

shreddedbacon:
Support for https://github.com/uselagoon/lagoon/pull/3709 and https://github.com/uselagoon/remote-controller/pull/267


bomoko (approved):
makes sense in line with the other changes

PR #730, opened 2024-12-22
fix: unauthenticated registry tls

shreddedbacon:
Small fix to the unauthenticated registry to configure the certificate properly for the ingress.


shreddedbacon:
This just fixes the unauthenticated registry so it gets the right certificate loaded up when not using harbor. Simple fix, still requires that users have set up the insecure registries option in docker https://docs.lagoon.sh/contributing-to-lagoon/developing-lagoon/#configure-docker

PR #744, opened 2025-01-23
fix: only use the configured hostname for opensearch output

smlx:
The reload_* options tell fluentd to ask the Opensearch Nodes API for
endpoint IP addresses to add to the round-robin pool for outbound
connections. This behaviour is never correct in our configuration
because we use TLS, and the certificate only has the configured
hostname, not the pod IPs.

So this change configures fluentd to only use the configured hostname.


tobybellwood (approved):
yes - simple enough change, fixes the issue!

PR #749, opened 2025-02-04
fix: role type for storage-calculator leader election

shreddedbacon:
Fixes the type for the role for storage-calculator from the wrong `ClusterRole` to `Role`


charts/lagoon-remote/Chart.yaml
```diff
@@ -19,7 +19,7 @@ type: application
 # This is the chart version. This version number should be incremented each
 # time you make changes to the chart and its templates, including the app
 # version.
-version: 0.96.2
+version: 0.96.3
```
tobybellwood, line 22: No need to bump the version in a PR :wink: 
```
suggestion
version: 0.96.2
```

shreddedbacon: yeah, i had every intention of this being its own release, but the other things got merged before

tobybellwood (approved):
simple enough, looks good to me!

PR #750, opened 2025-02-17
chore: update kind and tested k8s versions

tobybellwood:


.github/workflows/lint-test-matrix.yaml
```diff
@@ -11,12 +11,9 @@ jobs:
...
         kindest_node_version:
-        - v1.25.16@sha256:6110314339b3b44d10da7d27881849a87e092124afab5956f2e10ecdb463b025
-        - v1.26.15@sha256:1cc15d7b1edd2126ef051e359bf864f37bbcf1568e61be4d2ed1df7a3e87b354
-        - v1.27.17@sha256:3fd82731af34efe19cd54ea5c25e882985bafa2c9baefe14f8deab1737d9fabe
-        - v1.28.13@sha256:45d319897776e11167e4698f6b14938eb4d52eb381d9e3d7a9086c16c69a8110
-        - v1.29.8@sha256:d46b7aa29567e93b27f7531d258c372e829d7224b25e3fc6ffdefed12476d3aa
-        - v1.31.0@sha256:53df588e04085fd41ae12de0c3fe4c72f7013bba32a20e7325357a1ac94ba865
+        - v1.29.12@sha256:62c0672ba99a4afd7396512848d6fc382906b8f33349ae68fb1dbfe549f70dec
+        - v1.30.8@sha256:17cd608b3971338d9180b00776cb766c50d0a0b6b904ab4ff52fd3fc5c6369bf
+        - v1.32.0@sha256:c48c62eac5da28cdadcf560d1d8616cfa6783b58f0d94cf63ad1bf49600cb027
```
shreddedbacon, line 16: So we're only going to "support" 1.29+ now?

tobybellwood: kind 0.26.0 onwards only supports 1.29 onwards images (ie non-EOL) - so testing anything earlier would likely fail.

I don't think we should prohibit installation on <1.29 (yet) - as vendors can +1 year Kubernetes support dates, but maybe we might introduce a pre-release b/c test using an older kind or similar?

Makefile
```diff
@@ -343,7 +345,7 @@ install-k8upv2:
 		--set k8up.envVars[5].name=BACKUP_GLOBALRESTORES3BUCKET,k8up.envVars[5].value=baas-restores \
 		--set k8up.envVars[6].name=BACKUP_GLOBALRESTORES3ACCESSKEYID,k8up.envVars[6].value=lagoonFilesAccessKey \
 		--set k8up.envVars[7].name=BACKUP_GLOBALRESTORES3SECRETACCESSKEY,k8up.envVars[7].value=lagoonFilesSecretKey \
-		--version=4.8.2 \
+		--version=4.8.4 \
```
shreddedbacon, line 348: Also need to update the version in the CRD install step
https://github.com/uselagoon/lagoon-charts/blob/263cfad1632a442e9a5d202b2b7d4fb540b3daf4/Makefile#L332-L333

Maybe it is worth moving them to a variable, along with the others?

.github/workflows/lint-test-matrix.yaml
shreddedbacon: Where does it say that 1.29 is the minimum?

tobybellwood: https://github.com/kubernetes-sigs/kind/releases/tag/v0.26.0

shreddedbacon: Ah, its the minimum they supply as built to work with the release?

shreddedbacon (approved):
Easy. The fewer versions to test against in the lint test matrix is nice :)

PR #751, opened 2025-02-25
chore: use the lagoon-remote ssh-core token in tests

shreddedbacon:
Just a simple one that uses the the token created by the ssh core feature being enabled to be used instead of creating a short lived token.


tobybellwood (approved):
Another simple local dev change

PR #752, opened 2025-02-26
ssh and broker service settings

Jancis:
This change allows adding annotations, externalTrafficPolicy, loadBalancerSourceRanges for ssh and broker services. 

I'll do a chart version bump if this gets picked up, to avoid version clashes in case something else gets merged first.

### DIFF
--- a/charts/lagoon-core/Chart.yaml
+++ b/charts/lagoon-core/Chart.yaml
@@ -48,3 +48,5 @@ annotations:
       description: update lagoon-opensearch-sync image to v0.8.3
     - kind: changed
       description: update ssh-portal-api and ssh-token options
+    - kind: changed
+      description: ssh, ssh-token and broker service configuration options
--- a/charts/lagoon-core/templates/broker.service.yaml
+++ b/charts/lagoon-core/templates/broker.service.yaml
@@ -51,8 +51,19 @@ metadata:
   name: {{ include "lagoon-core.broker.fullname" . }}-amqp-ext
   labels:
     {{- include "lagoon-core.broker.labels" . | nindent 4 }}
+  {{- with .Values.broker.service.amqpExternal.annotations }}
+  annotations:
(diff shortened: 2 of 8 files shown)


shreddedbacon (changes requested):
There are other services that you may also want to extend with these options.

Those being the `ssh-token` [service](https://github.com/uselagoon/lagoon-charts/blob/main/charts/lagoon-core/templates/ssh-token.service.yaml) in `lagoon-core` and the `ssh-portal` [service](https://github.com/uselagoon/lagoon-charts/blob/main/charts/lagoon-remote/templates/ssh-portal.service.yaml) in `lagoon-remote`

Don't worry about chart version bumps, as we handle this when we do the releases. But please add a changelog entry to each `Chart.yml` in the [lagoon-core changelog](https://github.com/uselagoon/lagoon-charts/blob/main/charts/lagoon-core/Chart.yaml#L41-L50) and [lagoon-remote changelog](https://github.com/uselagoon/lagoon-charts/blob/main/charts/lagoon-remote/Chart.yaml#L41-L48) with the changes made too.

Edit: Clarification on version bump

Jancis:
Suggested changes implemented

shreddedbacon:
Thanks! It all looks good to me now. Just waiting for the tests to :heavy_check_mark: and we can look at merging.

shreddedbacon:
@Jancis sorry to bother, but would you be able to amend your commit and push again to trigger the test suites to run with the label I just added. I forgot that the `next-release` label was needed for this current release cycle so the tests are failing.

Jancis:
Sure! Let's see if this goes trough :crossed_fingers:

shreddedbacon (approved):
Thanks for the contribution @Jancis ! Will hopefully be part of the next release.

tobybellwood:
Yup, release is scheduled for next week, will get this in the queue!

PR #754, opened 2025-02-28
test: label checks

shreddedbacon:


.github/workflows/test-suite.yaml
```diff
@@ -131,16 +153,14 @@ jobs:
...
-        (contains(github.event.pull_request.labels.*.name, 'needs-testing'))
+        (steps.list-changed.outputs.changed == 'true' && !needs.get-labels.outputs.nextRelease == 'next-release') ||
+        (needs.get-labels.outputs.needsTesting == 'needs-testing')
       run: |
         make install-lagoon
         make -j8 -O fill-test-ci-values TESTS=[${{ matrix.test }}]
 
     - name: Helm-install the test fixtures and fill lagoon-test/ci/linter-values.yaml (next-release)
-      if: |
-        (steps.list-changed.outputs.changed == 'true') &&
```
tobybellwood, line 142: this one might need to stay in to skip the tests if nothing has changed, but convert to an `||` instead

shreddedbacon: yeah ill make it an `||`

shreddedbacon: The previous logic was "if changed & label exists" but that wouldn't fire if just the label was added or removed. i think it still makes sense to leave it as "if label exits" and ignore the "if changed" so it runs in both conditions

CGoodwin90 (approved):

PR #760, opened 2025-03-13
chore: removes workflows

bomoko:
This PR removes the workflow subsystem that is slated to be deprecated in Lagoon 2.25.0


charts/lagoon-core/Chart.yaml
```diff
@@ -54,3 +54,5 @@ annotations:
       description: ssh, ssh-token and broker service configuration options
     - kind: changed
       description: update nats to v1.2.11
+    - kind: changed
+      description: Lagoon Workflows subsystem removed
```
shreddedbacon, line 58: Since we changed to the model of merging to main and then doing a release of the charts as a separate version bump. You'll need to remove all the previous changelog entries from this pullrequest, since all the other changes were part of chart `v1.51.0`. The same applies to the `lagoon-test` chart yaml.

charts/lagoon-core/Chart.yaml
```diff
@@ -42,3 +42,5 @@ annotations:
   artifacthub.io/changes: |
     - kind: changed
       description: update lagoon appVersion to v2.24.1
```
shreddedbacon, line 44: This can be removed, since 2.24.1 has been released now.

The changes in this PR would effectively be the start of the next release.

shreddedbacon: Same for the `lagoon-test` chart yaml.

shreddedbacon (changes requested):

charts/lagoon-core/Chart.yaml
bomoko: cool - done.

shreddedbacon (approved):
Great, lets get this in.

PR #762, opened 2025-03-20
fix: storage-calculator cluster role

shreddedbacon:
After the secret changes were added to storage calculator, the role wasn't updated to reflect the requests to get the secret. This fixes the role so that the calculator can retrieve secrets.


tobybellwood (approved):
Makes sense!

PR #763, opened 2025-03-26
chore: fix missing backup-handler image override

shreddedbacon:
While testing something with k8up in a local-stack, the backup-handler image fails to pull if checked out in a working branch. This fixes it so that it correctly uses the built and pushed image if not using a stable local-stack.


tobybellwood (approved):
LGTM

PR #767, opened 2025-04-01
chore: support logs2notifications options

shreddedbacon:
Add make variables to enable or disable certain notifications, also enables slack by default


tobybellwood (approved):
Simple change for local dev

PR #769, opened 2025-04-02
chore: add dedicated label to docker-host services

shreddedbacon:
To make selecting from the individual docker hosts with their direct services, this adds a dedicated label to be able to select only the direct services in the cluster.


tobybellwood (approved):
LGTM

PR #773, opened 2025-05-06
Add startup probe and allow configuration of probes

anothertobi:
With a full buffer the logs concentrator can take longer to start up
than the default liveness probe. Making the probes configurable allows
operators to adjust them as required according to the size of the
environment.


smlx (approved):
Makes probes configurable, and adds a default 100 second startup time to the fluent container.

LGTM, thanks!

PR #774, opened 2025-05-06
chore: Adds extra env args to insights remote

bomoko:
This PR adds the ability to add extra environment variables to insights remote, in preparation for some integrations it's enabling.


shreddedbacon (approved):
Pretty straightforward

PR #782, opened 2025-05-08
test: disable tls in lagoon-build-deploy for older stable core

shreddedbacon:
Fixes a small issue where the `lagoon-build-deploy` chart which is now configured in CI to install with broker tls enabled but has been released. When the `lagoon-core` installation in the uselagoon/lagoon repo installs the stable version prior to the upcoming release which contains broker enabled by default, the `lagoon-build-deploy` chart won't connect.

This forces the lagoon-build-deploy chart to install without broker tls if the version of lagoon-core is older than 1.52.0 which is the last version that supported non-tls installations in CI by default

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,6 +88,12 @@ STABLE_CORE_CHART_VERSION =
 STABLE_REMOTE_CHART_VERSION = 
 STABLE_BUILDDEPLOY_CHART_VERSION = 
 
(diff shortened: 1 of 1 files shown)


rocketeerbkw (approved):
Tested locally with `make k3d/stable-local-stack` and TLS is disabled

PR #795, opened 2025-06-18
Adds a. prefix as a fix for nip.io resolution

bomoko:
Nip.io parsing of urls has changed since being moved to sslip.io

In order to fix the testing framework, we've made a small change to the testing domains (adding a `.a.` prefix) to ensure that our ip addresses are correctly parsed.


shreddedbacon (approved):
Nice and simple work around for the issue

PR #796, opened 2025-06-26
chore: use k8up.io/v1 schedule if available

### LINKED ISSUE #707: lagoon-core: use k8up.io/v1 for backup schedule
Currently we only support `backup.appuio.ch/v1alpha1`, if `k8up.io/v1` is detected it should be used instead.

I did [this](https://github.com/uselagoon/lagoon-charts/compare/k8up-version), there is probably a better way if anyone else wants to have a crack at it.

shreddedbacon:
Since k8up.io (k8up 2.x) is supported by `lagoon-remote` and `lagoon-build-deploy` for workloads deployed by Lagoon, it makes sense to also have the chart for lagoon-core support k8up.io

Closes #707

### DIFF
--- a/charts/lagoon-core/templates/k8up.schedule.yaml
+++ b/charts/lagoon-core/templates/k8up.schedule.yaml
@@ -1,3 +1,34 @@
+{{- if .Capabilities.APIVersions.Has "k8up.io/v1/Schedule" }}
+{{- $schedule := index (lookup "k8up.io/v1" "Schedule" .Release.Namespace (include "lagoon-core.fullname" . )) | default dict }}
+{{- $bucket := coalesce .Values.k8upBackupBucketName (dig "spec" "backend" "s3" "bucket" "" $schedule) (print "baas-" (include "lagoon-core.fullname" .) "-" (randAlphaNum 8 | lower)) }}
+apiVersion: k8up.io/v1
+kind: Schedule
+metadata:
+  name: {{ include "lagoon-core.fullname" . }}
+  labels:
+    {{- include "lagoon-core.labels" . | nindent 4 }}
+spec:
+  backend:
+    repoPasswordSecretRef:
+      key: repo-pw
+      name: "{{ include "lagoon-core.fullname" . }}-baas-repo-pw"
+    s3:
+      bucket: {{ $bucket | quote }}
+      {{ with .Values.k8upS3Endpoint }}
+      endpoint: {{ . | quote }}
+      {{ end }}
+  backup:
+    schedule: '15 * * * *'
+  check:
+    schedule: '45 3 * * *'
+  prune:
+    retention:
+      keepHourly: 36
+      keepDaily: 30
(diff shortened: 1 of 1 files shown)


rocketeerbkw (commented):
I tried testing this locally and I don't think it worked. I used `make/k3d-localstack` to install k8up v1 and v2 which created a v1 schedule. I noted the randomly generated bucket name. Then I re-installed lagoon-core with this PR and I got a new v2 schedule with a different randomly generated bucket name. And also the v1 schedule still exists.

It could be that my testing order is incorrect though.

shreddedbacon:
Made some adjustments to the k8up installation process to do some more testing with this.

You can use the following to setup a local-stack with the specific charts branches that contain the fixes, starting from this branch https://github.com/uselagoon/lagoon/pull/4015

```
# in `uselagoon/lagoon` repository, checkout the upstream branch for the proper flag support
git checkout local-stack-k8up-adjustments
# then run the make targets to setup the cluster
make k3d/local-stack INSTALL_K8UP=true INSTALL_K8UP_VERSIONS=v1,v2 CHARTS_TREEISH=local-stack-k8up-adjustments
```
Once that is done, you'll be able to check that only the `backup.appuio.ch` resource is created

Then checkout the k8up-version charts and run the `install-lagoon` only target. I've left the flags for k8up and treeish on the install target mostly to be clear, but they're not used.
```
make k3d/checkout-charts CHARTS_TREEISH=k8up-version
make k3d/install-lagoon INSTALL_K8UP=true INSTALL_K8UP_VERSIONS=v1,v2 CHARTS_TREEISH=k8up-version
```
You'll now have both a `backup.appuio.ch` schedule, and the new `k8up.io` schedule. The new schedule should now contain the correct bucket.

I'm not sure how to clean up the older schedule though :sweat_smile:

Deleting the `backup.appuio.ch` schedule and then re-running the install still retains the correct bucket in the new schedule, and the old one doesn't get recreated. I'm not sure if we should try to figure out how to get helm to then also clean up the older resource, or just call it out as a step in the release notes.

rocketeerbkw:
Retaining the backup bucket name is working now.

> I'm not sure if we should try to figure out how to get helm to then also clean up the older resource, or just call it out as a step in the release notes.

I guess the only automatic option is to add a chart hook? I think calling it out in the release notes is enough unless there is some risk in breaking backups when both schedules exist.

shreddedbacon:
> I guess the only automatic option is to add a chart hook? I think calling it out in the release notes is enough unless there is some risk in breaking backups when both schedules exist.

Only one of them will succeed due to locking that takes place. I think release notes is fine, just have to be very explicit in calling it out I guess. I might update this to put something into the [NOTES.txt](https://github.com/uselagoon/lagoon-charts/blob/main/charts/lagoon-core/templates/NOTES.txt) file that if the k8up change takes place, a note is left in the resulting output.

rocketeerbkw (approved):
LGTM

PR #802, opened 2025-07-08
feat: seeded user and org

jackwrfuller:
## Lagoon Core Seeded User and Organisation

Addresses https://github.com/uselagoon/lagoon/issues/3942.

This PR introduces an optional post-install and post-upgrade helm chart hook for Lagoon Core that enables an initial user and/or an initial organization to be created automatically, with the details configurable as below.

This feature is enabled by setting the following in the chart's `values.yml`:
```yaml
lagoonSeedUsername: test@example.com
lagoonSeedPassword: password
lagoonSeedOrganization: example-org
```

Either `lagoonSeedOrganization` or both `lagoonSeedPassword` and `lagoonSeedUsername` must be set for the hook to run.


charts/lagoon-core/files/seed-core.py
shreddedbacon, line 1: We usually contain the business logic for this sort of thing in the main `uselagoon/lagoon` repository, this way we could potentially leverage existing code, or write it in a different language.

Similar to how we do for the api database migration and broker bootstrap jobs
* https://github.com/uselagoon/lagoon-charts/blob/main/charts/lagoon-core/templates/api.migratedb.job.yaml#L21-L30
* https://github.com/uselagoon/lagoon-charts/blob/main/charts/lagoon-core/templates/broker.bootstrap.job.yaml#L22-L29

For now, we can probably go with this being defined in the charts, unless there are other objections to it.

Apart from that, if I only wanted to create the initial organization with no user, this would not work.

charts/lagoon-core/templates/seed-core.job.yaml
```diff
@@ -0,0 +1,48 @@
...
+  name: "lagoon-core-seeding"
+  annotations:
+    "helm.sh/hook": post-install,post-upgrade
+spec:
+  template:
+    spec:
+      restartPolicy: Never
+      containers:
+        - name: seed-core-job
+          image: "uselagoon/tests"
```
shreddedbacon, line 14: We wouldn't typically use this image in a production job. As I said in my other comment, we would usually execute a command in another image that would be classed as production worthy. The API image might be able to be used here, and leverage the same logic for `image` tags as [other jobs ](https://github.com/uselagoon/lagoon-charts/blob/main/charts/lagoon-core/templates/api.migratedb.job.yaml#L21-L30)

charts/lagoon-core/templates/seed-core.job.yaml
```diff
@@ -0,0 +1,48 @@
...
+          - name: JWTSECRET
+            valueFrom:
+              secretKeyRef:
+                name: lagoon-core-secrets
+                key: JWTSECRET
+          - name: KEYCLOAK_ADMIN_API_CLIENT_SECRET
+            valueFrom:
+              secretKeyRef:
+                name: lagoon-core-keycloak
+                key: KEYCLOAK_ADMIN_API_CLIENT_SECRET
```
shreddedbacon, line 43: You'll need to fix up the secretKeyRef to use the helm sourced name like [other jobs](https://github.com/uselagoon/lagoon-charts/blob/main/charts/lagoon-core/templates/broker.bootstrap.job.yaml#L36)

jackwrfuller: Changed to use templated refs.

charts/lagoon-core/templates/seed-core.job.yaml
jackwrfuller: Modified to use the api image, with the script now being js run by the api's nodejs.

charts/lagoon-core/files/seed-core.py
jackwrfuller: For the nonce, I've kept the script as a file in the chart but it would be trivial to transfer it into the main Lagoon repo later if we feel it is better there.

I have modified the script so that you can create an org without a user.

jackwrfuller:
> There are also linting issues to resolve in the `values.yaml` file.
> 
> Edit: It may be worth reviewing the other jobs to see if there are other helm template changes that you may need to correct in your job template, like labels/annotations, etc.

I believe I have addressed the linting issues, but I'm not sure - I'm finding the CI hard to parse. Is there any docs for what tests the CI runs?

charts/lagoon-core/Chart.yaml
```diff
@@ -40,9 +40,5 @@ dependencies:
...
 annotations:
   artifacthub.io/changes: |
-    - kind: changed
-      description: add support for s3 files bucket to actions-handler
-    - kind: changed
-      description: add servicemonitor for keycloak
-    - kind: fixed
-      description: support for s3 files bucket to actions-handler
+    - kind: added
+      description: add support for setting seed user or organization upon initial install
```
shreddedbacon, line 44: You'll need to add all the ones you deleted back, as they are part of the yet to be released chart version

jackwrfuller: Done. I was under the mistaken assumption that each new PR replaced this section entirely, since that was what i noticed happening in the PRs I inspected. That makes a lot more sense though.

shreddedbacon: Yeah, it depends on the changes. It is a bit chaotic to manage at the moment, trying to think of a better way to make it clear if its released/unreleased change log items.

jackwrfuller:
PR is failing the test-suite (api) test. After spending some time reading the logs, there doesnt appear to be any issues with the seeding pod, rather some seemingly (at least to me) unrelated ones. Before I sink too much more time debugging this, are you able to confirm I should indeed be passing this test?

charts/lagoon-core/files/seed-core.js
```diff
@@ -0,0 +1,206 @@
...
+
+    AddUserToOrg: addAdminToOrganization(input: {
+        user: { email: "${INITIAL_USER_EMAIL}" }
+        organization: { id: ${orgId} }
+        role: OWNER
+    }) {
+        id
+    }
+}
+`;
```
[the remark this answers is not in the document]
jackwrfuller: So just remove entirely do you think?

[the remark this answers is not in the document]
shreddedbacon: Yep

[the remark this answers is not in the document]
jackwrfuller: okay, removed.

jackwrfuller:
> You'll also need to update this with the latest changes from main

Done. 

I lowkey want to rename `initialUserName` to `lagoonSeedUsername` to reflect your makefile suggestions, I think your naming is way better.

shreddedbacon:
> I lowkey want to rename `initialUserName` to `lagoonSeedUsername` (and the others) to reflect your makefile suggestions, I think your naming is way better.

I'm fine with either

charts/lagoon-core/files/seed-core.js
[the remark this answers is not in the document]
jackwrfuller: I believe I fixed it by changing the import syntax to

```js
const axios = require("axios");
const jwt = require("jsonwebtoken");
```

Screenshot 2025-07-24 at 9 30 15 am

Makefile
```diff
@@ -605,6 +610,9 @@ endif
 		$$([ $(LAGOON_CORE_USE_HTTPS) = true ] && echo '--set broker.ingress.tls[0].secretName=broker-tls') \
 		$$([ $(LAGOON_CORE_USE_HTTPS) = true ] && echo '--set-string broker.ingress.annotations.kubernetes\\.io/tls-acme=true') \
 		$$([ $(LAGOON_CORE_USE_HTTPS) = true ] && echo '--set-string broker.ingress.annotations.nginx\\.ingress\\.kubernetes\\.io/ssl-redirect=false') \
+		$$([ $(LAGOON_SEED_USERNAME) ] && echo '--set lagoonSeedUsername=$(LAGOON_SEED_USERNAME)') \
+        $$([ $(LAGOON_SEED_PASSWORD) ] && echo '--set lagoonSeedPassword=$(LAGOON_SEED_PASSWORD)') \
+        $$([ $(LAGOON_SEED_ORGANIZATION) ] && echo '--set lagoonSeedOrganization=$(LAGOON_SEED_ORGANIZATION)') \
```
jackwrfuller, line 615: Screenshot 2025-07-24 at 9 41 35 am
 
Im getting stitched up here....

charts/lagoon-core/Chart.yaml
```diff
@@ -21,7 +21,7 @@ type: application
 # time you make changes to the chart and its templates, including the app
 # version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.54.2
+version: 1.54.3
```
shreddedbacon, line 24: Can you revert this please, we'll handle bumping the version as we close out the release cycle.

shreddedbacon (changes requested):

shreddedbacon:
@bomoko just looping you in for a second set of :eyes: 

I'm pretty happy with this as it is though.

charts/lagoon-core/files/seed-core.js
```diff
@@ -0,0 +1,167 @@
...
+  const lagoonApiUrl = LAGOON_API_URL.replace("https://", "http://");
+  const headers = {
+    "Content-Type": "application/json",
+    Authorization: `Bearer ${TOKEN}`
+  };
+
+  try {
+    const response = await axios.post(lagoonApiUrl, { query: seedOrgGql }, { headers });
+    const body = response.data;
+    console.log(body);
```
bomoko, line 65: Is this debug logging? I see the body var is also logged in the "no org id" conditional.

bomoko (commented):
This all looks good to me. Happy to approve if you're good @shreddedbacon

charts/lagoon-core/files/seed-core.js
shreddedbacon: Yeah, they definitely look like debug logging, can probably be removed for sure

jackwrfuller: I left the logging in so that the person install lagoon could debug the job if whatever reason it ended up not working.

shreddedbacon: In that case, maybe its better to put the log if it errors?

jackwrfuller: yeah agreed. I think theres two kinds of things being logged here: status updates and debug statements. I think the former are fine to keep (e.g L81) but the latter should only be included on error.

jackwrfuller: Removed print debug statements. I'll mark as resolved, but we can reopen if you think more should be removed.

shreddedbacon: I think for now, all good. It is an optional job as it is, and really will only be used on fresh installations.

shreddedbacon (approved):

PR #807, opened 2025-07-15
Pass extraEnvs to environment variables directly from values

dimiechatz:
Adds support of `valueFrom` field when adding environment variables in a container by passing `extraEnvs` to environment variables directly from values.

Related issue: https://amazeeio.atlassian.net/browse/PLAT-571

### DIFF
--- a/charts/lagoon-build-deploy/Chart.yaml
+++ b/charts/lagoon-build-deploy/Chart.yaml
@@ -16,14 +16,14 @@ kubeVersion: ">= 1.25.0-0"
 
 type: application
 
-version: 0.36.0
+version: 0.35.0
 
 appVersion: v0.24.0
 
 annotations:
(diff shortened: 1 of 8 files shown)


charts/lagoon-build-deploy/Chart.yaml
```diff
@@ -16,14 +16,14 @@ kubeVersion: ">= 1.25.0-0"
 
 type: application
 
-version: 0.35.0
+version: 0.36.0
```
dimiechatz, line 19: @bomoko not sure if we want to increase the minor or patch version for this change. Since this wasn't a bug, I've bumped the minor one.

dimiechatz: No need for bumping the version in the end.

PR #811, opened 2025-07-22
chore: add default certificate for ingress

shreddedbacon:
Configures a default wildcard certificate to ingress-nginx for local development


bomoko (approved):
makes perfect sense. Template out the cert for the ingress using the standard pattern. Looks good.

PR #821, opened 2025-08-05
Make logging-operator images configurable

anothertobi:
In some setups it can be required to use an alternative container image
registry than those used by the operator.


tobybellwood (approved):
yup - this makes sense!

PR #824, opened 2025-08-12
feature: Use public CA for lagoon-logs-concentrator's cert

dimiechatz:
Related issue: https://amazeeio.atlassian.net/browse/PLAT-547

## Changes
- Introduce `mTLS` value to allow usage of a cert from a public CA
- Fixes volume names for opensearch cert

### DIFF
--- a/charts/lagoon-logging/Chart.yaml
+++ b/charts/lagoon-logging/Chart.yaml
@@ -19,7 +19,7 @@ type: application
 # time you make changes to the chart and its templates, including the app
 # version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.88.0
+version: 0.89.0
 
 # AppVersion is set here the same as the logging-operator chart version to
 # autopopulate the post-install CRD message.
--- a/charts/lagoon-logs-concentrator/Chart.yaml
+++ b/charts/lagoon-logs-concentrator/Chart.yaml
@@ -19,7 +19,7 @@ type: application
 # time you make changes to the chart and its templates, including the app
 # version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.53.0
+version: 0.54.0
 
 # This section is used to collect a changelog for artifacthub.io
 # It should be started afresh for each release


charts/lagoon-logs-concentrator/templates/fluent-conf.configmap.yaml
```diff
@@ -33,10 +33,16 @@ data:
...
+        {{- if .Values.mTLS }}
         ca_path /fluentd/tls/ca.crt
         cert_path /fluentd/tls/server.crt
         private_key_path /fluentd/tls/server.key
         client_cert_auth true
+        {{- else }}
+        cert_path /fluentd/tls/tls.crt
+        private_key_path /fluentd/tls/tls.key
+        client_cert_auth false
+        {{- end }}
```
anothertobi, line 45: With mTLS disabled there won't be a certificate to use. It might still required the `ca_path` to be set, but the docs are not clear on whether the system default will be used or not (https://docs.fluentd.org/configuration/transport-section).

anothertobi (commented):
non-blocking: There are also some seemingly unrelated workflow pinning changes in the PR which might be unwanted.

charts/lagoon-logs-concentrator/templates/fluent-conf.configmap.yaml
dimiechatz: In the statefulset we're mounting the secret/cert which will be created by cert-manager in this case: https://github.com/uselagoon/lagoon-charts/blob/main/charts/lagoon-logs-concentrator/templates/statefulset.yaml#L87

From the fluentd docs for public CAs,  it isn't clear if  `ca_path `, `cert_path` and `private_key_path` need to be set: https://docs.fluentd.org/configuration/transport-section#signed-public-ca-parameters.

dimiechatz: Indeed, after testing `ca_path` doesn't have to be set. However, we still need to set the `cert_path`. Othewise, fluentd is failing to start with:
```bash
[error]: config error file="/fluentd/etc/fluent.conf" error_class=Fluent::ConfigError error="no valid cert options configured. specify either 'cert_path', 'ca_cert_path' or 'insecure'"
```

and if we set the `ca_cert_path` only, it will ask for `private_key_path`.

anothertobi (approved):

PR #827, opened 2025-08-26
fix: local-stack installation of k8up on ARM systems

jackwrfuller:
# Description

By default, the local-stack installs k8upv1alpha1 ("v1"), but this does not support ARM systems. This makes it impossible to build the local-stack directly with backups enabled using `make k3d/local-stack INSTALL_K8UP=true` without manually removing the offending make target dependency.

This PR changes the installation so that v1 is only installed if explicitly enabled via the `REMOTE_CONTROLLER_K8UP_VERSION` variable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -489,7 +489,7 @@ endif
 # install k8up versions for backup upgrade path verifications if requested
 ifeq ($(INSTALL_K8UP),true)
 ifeq ($(REMOTE_CONTROLLER_K8UP_VERSION), v1)
-install-lagoon-dependencies: install-k8upv1
+install-lagoon-dependencies: install-k8upv1 install-k8upv2
 else
 install-lagoon-dependencies: install-k8upv2
 endif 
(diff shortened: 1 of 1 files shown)


shreddedbacon (approved):

shreddedbacon:
The basic idea here is good, and I've approved it generally.

It would still be good to be able to install k8upv1 though optionally along side v2 for testing migration paths for those that aren't using arm based systems.

shreddedbacon (changes requested):
I've changed my mind, can you please refactor this to support the installation of both versions some how.

jackwrfuller:
Based on conversation, I have refactored so that setting `REMOTE_CONTROLLER_K8UP_VERSION` to "v1" installs both v1 and v2.

shreddedbacon (approved):
Yep, this works. ARM users just won't be able to do migration tests for v1 to v2.

PR #829, opened 2025-09-03
chore: remove bitnami charts

shreddedbacon:
Since Bitnami changed all their stuff around, it has the potential for flow on issues for our charts and local testing. This removes all bitnami references and uses some charts that are relatively new that are mostly drop in replacements, but the key difference is they use upstream images from the providers, rather than bitnami images.

Cloudpirates.io has made charts available basically as bitnami alternatives, see their blog post https://www.cloudpirates.io/knowledge/blog/unsere-open-source-helm-charts-als-bitnami-alternative

As these charts are only used for CI/local development purposes, this is more than suitable for now.

Bonus, it looks like this allows mongodb to be used on ARM systems now as the [upstream mongo](https://hub.docker.com/_/mongo) image appears to provide this.

### DIFF
--- a/.github/workflows/test-suite.yaml
+++ b/.github/workflows/test-suite.yaml
@@ -107,7 +107,6 @@ jobs:
         helm repo add harbor https://helm.goharbor.io
         helm repo add ingress-nginx https://kubernetes.github.io/ingress-nginx
         helm repo add stable https://charts.helm.sh/stable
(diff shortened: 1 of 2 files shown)


bomoko (approved):

shreddedbacon:
Just rebased for merge requirements

PR #837, opened 2025-09-16
chore: retry apt install due to locking

shreddedbacon:
Implement a retry logic around the apt install dependency to prevent the following from failing actions
```
E: Could not get lock /var/lib/dpkg/lock-frontend. It is held by process 2358 (apt-get)
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?
```

### DIFF
--- a/.github/workflows/test-suite.yaml
+++ b/.github/workflows/test-suite.yaml
@@ -44,7 +44,12 @@ jobs:
         fetch-depth: "0"
 
     - name: Set up chart-testing dependencies
-      run: sudo apt-get -y install python3-wheel
+      run: |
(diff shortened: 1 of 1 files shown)


tobybellwood (approved):
looks fine to me - there's also a good chance that wheel is already installed, but this will do for now

```
Run for i in $(seq 1 10); do
Reading package lists...
Building dependency tree...
Reading state information...
python3-wheel is already the newest version (0.42.0-2).
python3-wheel set to manually installed.
0 upgraded, 0 newly installed, 0 to remove and 23 not upgraded.
```

PR #842, opened 2025-09-20
fix: seeding job api references and organization name setup

shreddedbacon:
Just some small adjustments to the seed user/org job. 

First is to set the organization name to be made "safe" and lowercase. While setting the full value as the friendly name.

Second adjusts the keycloak url to use the internal one if frontendurl is not provided, and prefer to use the internal lagoon api endpoint

### DIFF
--- a/charts/lagoon-core/Chart.yaml
+++ b/charts/lagoon-core/Chart.yaml
@@ -50,3 +50,7 @@ annotations:
       description: update opensearch-sync to v0.9.1
     - kind: changed
       description: update insights-handler to v0.0.9
+    - kind: fixed
+      description: fix seeding job to use internal service names for connecting to api
+    - kind: fixed
+      description: fix seeding job to set safe organization name and friendlyname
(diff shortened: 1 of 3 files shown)


shreddedbacon:
@jackwrfuller this is where I got hung up with some things with the seed job

shreddedbacon:
@jackwrfuller if you can review this, that'd be great. Should mostly be a :white_check_mark: and flick.

I changed the template to use just the internal keycloak record, since this job runs within the namespace.

Running in a local-stack, everything works as expected.
```
Seeding Lagoon core with initial user
Connected to keycloak
Setting user password in keycloak at http://lagoon-core-keycloak:8080/auth
Keycloak token successfully retrieved
Password set successfully!
Seeding Lagoon core with initial organization
Created org with id 1
```

If trying to seed an organization name with spaces or special characters using the local-stack, it fails because of the way we pass the variables through. This doesn't occur when using the chart values directly though. Not worth fixing this though, as its a local-stack problem only.

For example, If I set this in the values directly, it creates the organization correctly in Lagoon using the seeding.
```
lagoonSeedOrganization: "Bob's Burgers"
```
It fails in the local-stack though if I use `LAGOON_SEED_ORGANIZATION="Bob's Burgers"` in the make variables.

jackwrfuller (approved):
Seems fine to me. 

Tested in local-stack and I'm getting the same output, and can reconfirm user and org are created as intended.

PR #843, opened 2025-09-20
refactor: configuration option to disable external nats concentrator service

shreddedbacon:
The nats concentrator service is created by default. In cases where this isn't required, it should be able to be disabled.

This adds the option to disable it if required, the default remains enabled.


shreddedbacon:
@jackwrfuller another quick one that should mostly be a :white_check_mark: and flick.

The behaviour remains the same, but allows for disabling the service in a situation where you don't need to expose the nats concentrator, like where core+remote are installed in the same cluster. This doesn't need an external concentrator service.

jackwrfuller (approved):
Tested in local-stack, and I can confirm setting `natsService.enabled=false` does indeed remove the `lagoon-core-nats-concentrator` service.

So besides the following question, it looks good to me. 

Should we disable this service by default in the local-stack?

shreddedbacon:
> Should we disable this service by default in the local-stack?

I think its fine to leave it, it could be useful for testing. It doesn't have any real impact in the local-stack by remaining.

But for an actual deployment in a cloud, it would result in a loadbalancer being created and potentially have a cost associated to it, even if not being used.

PR #858, opened 2025-11-19
feat: use new webhook-handler

shreddedbacon:
Updates charts to support new webhook-handler and removes unused webhooks2tasks. Required for https://github.com/uselagoon/lagoon/pull/3979


rocketeerbkw (approved):
Tested working locally 👍

PR #906, opened 2026-04-21
chore: remove leftover harbor configs in lagoon-core

rocketeerbkw:
Harbor support in Lagoon core was removed in 2.20.x, this PR just removes leftover helm config, including required values that end up polluting values files with things like `harborAdminPassword: not-needed`.

Not bumping chart version since I expect other PRs to do that for the next release.


shreddedbacon (approved):
Ran through in a local-stack and everything working as expected still.

PR #909, opened 2026-05-05
fix(lagoon-logging): review kubernetes_metadata filters to reduce API 404 rate

psh396:
### Problem
The kubernetes_metadata Fluentd plugin enriches log records with namespace labels by making pod API calls (GET /api/v1/namespaces/{ns}/pods/nopod). The nopod pod name is a deliberate dummy used to trigger a namespace lookup as a side effect, but it generates a high volume of 404 audit events on every cluster this chart is deployed to, creating noise in audit logs and unnecessary API server load.

https://grafana.test6.amazee.io/goto/bsWgnj1DR?orgId=1

cloudwatch logs https://eu-central-1.console.aws.amazon.com/cloudwatch/home?region=eu-central-1#log-analytics?active=a&a.id=ca4b646b-700c-4d68-800e-4dba4a1ceff3&a.pos=0&a.label=Query+1&a.type=query&a.query=SOURCE+START%3D-259200s+END%3D0s+%7C%0Afields+%40timestamp%2C+%40message%0A%7C+filter+%40message+like+%2F%5Cs4%5Cs%2F%0A%7C+filterIndex+%40data_source_type+in+%5B%22audit%22%2C+%22api_server%22%2C+%22authenticator%22%2C+%22controller_manager%22%2C+%22Unknown%22%2C+%22scheduler%22%5D%0A%7C+sort+%40timestamp+asc

### Solution

Apply the recommendations from the [AWS EKS Fluentd at-scale guide](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html):

Set watch false to eliminate LIST/WATCH calls to the API server
Set skip_pod_lookup true to prevent pod-level API lookups while still fetching namespace labels (required for index routing via lagoon.sh/project and lagoon.sh/environmentType)

### Changes

- Add watch false
- Skip pod lookup

Closes AIOP-5411

### DIFF
--- a/charts/lagoon-logging/templates/logs-dispatcher.fluent-conf.configmap.yaml
+++ b/charts/lagoon-logging/templates/logs-dispatcher.fluent-conf.configmap.yaml
@@ -157,6 +157,7 @@ data:
       @log_level warn
       skip_container_metadata true
       skip_master_url         true
+      cache_ttl {{ default 3600 .Values.logsDispatcher.kubernetesMetadataCacheTTL }}
     </filter>
     # strip the duplicate information so that it doesn't appear in logs
     <filter lagoon.*.container>
@@ -208,6 +209,7 @@ data:
       @log_level warn
       skip_container_metadata true
       skip_master_url         true
+      cache_ttl {{ default 3600 .Values.logsDispatcher.kubernetesMetadataCacheTTL }}
     </filter>
     # strip the dummy information so that it doesn't appear in logs
     <filter lagoon.*.application>
@@ -287,6 +289,7 @@ data:
         @log_level warn
         skip_container_metadata true
         skip_master_url         true
+        cache_ttl {{ default 3600 .Values.logsDispatcher.kubernetesMetadataCacheTTL }}
       </filter>
       # strip the dummy information so that it doesn't appear in logs
       <filter lagoon.*.router.nginx>
@@ -324,6 +327,7 @@ data:
       @log_level warn
       skip_container_metadata true
       skip_master_url         true
+      cache_ttl {{ default 3600 .Values.logsDispatcher.kubernetesMetadataCacheTTL }}
     </filter>
 
 {{- end }}
--- a/charts/lagoon-logging/templates/logs-dispatcher.fluent-conf.configmap.yaml
+++ b/charts/lagoon-logging/templates/logs-dispatcher.fluent-conf.configmap.yaml
@@ -211,11 +211,13 @@ data:
     </filter>
     # enrich with k8s metadata (will get the namespace labels)
     # watch disabled and pod labels skipped - see container filter comment above.
+    # skip_pod_lookup is set because the pod_name above is synthetic (the
+    # application host) and would cause API 404s.
     <filter lagoon.*.application>
       @type kubernetes_metadata
       @log_level warn
       watch                   false
-      skip_pod_labels         true
+      skip_pod_lookup         true
       skip_container_metadata true
       skip_master_url         true
     </filter>
@@ -293,11 +295,13 @@ data:
       </filter>
       # enrich with k8s metadata (will get the namespace labels)
       # watch disabled and pod labels skipped - see container filter comment above.
+      # skip_pod_lookup is set because the pod_name above is 'nopod' (synthetic)
+      # and would cause API 404s.
       <filter lagoon.*.router.nginx>
         @type kubernetes_metadata
         @log_level warn
         watch                   false
-        skip_pod_labels         true
+        skip_pod_lookup         true
         skip_container_metadata true
         skip_master_url         true
       </filter>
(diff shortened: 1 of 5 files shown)


dimiechatz (commented):
Before we merge the changes, could you test/confirm that the issue is actually being solved in a test or internal cluster?

charts/lagoon-logging/values.yaml
```diff
@@ -86,6 +86,10 @@ logsDispatcher:
 
   extraEnvsFrom: []
 
+  # TTL in seconds for the kubernetes_metadata plugin cache (covers both
+  # successful and failed/404 lookups). Increase to reduce API server load.
+  kubernetesMetadataCacheTTL: 3600
```
dimiechatz: If 1hr is the default, we should increase it in the helm-catalog MR.

psh396: yes, which value should I do in helm?

charts/lagoon-logging/templates/logs-dispatcher.fluent-conf.configmap.yaml
```diff
@@ -281,6 +281,23 @@ data:
           _dummy_ ${record['kubernetes'] = {'namespace_name' => record['namespace'], 'pod_name' => 'nopod', 'container_name' => 'nocontainer'}; record['docker'] = {'container_id' => "#{record['namespace']}_#{record['ingress_name']}"}; nil}
```
psh396, line 293: I am not sure if we should substitute this expression, or leave 2 expressions, 2 expressions shouldnt increase latency

psh396:
sorry didnt have time to roll out this to test clusters today

charts/lagoon-logging/templates/logs-dispatcher.fluent-conf.configmap.yaml
```diff
@@ -281,6 +281,23 @@ data:
...
+      # Normalise the fake docker container_id to a per-namespace constant.
+      # The snippet above derives the id from ingress_name, giving one cache
+      # entry per (namespace, ingress_name) pair in @id_cache. In a cluster
+      # with many ingress routes this produces a large number of unique ids,
+      # each triggering its own GET …/pods/nopod → 404 on first occurrence.
+      # Replacing the id with a fixed per-namespace string means at most ONE
+      # 404 API call per namespace; every subsequent record is a cache hit.
+      # (Removing docker entirely would break enrichment: lookup_from_k8s_field
+      # requires both kubernetes and docker keys to be present and returns the
+      # record unchanged when docker is absent.)
```
dimiechatz:
```suggestion
      # Normalize the fake docker container_id to a per-namespace constant.
      #
      # Context: Deriving the ID from the ingress_name creates a unique cache entry
      # for every route, triggering a 404 K8s API call (GET .../pods/nopod) for each one.
      #
      # Fix: Using a fixed per-namespace string reduces this to exactly ONE 404 call
      # per namespace. All subsequent records will safely hit the cache.
      #
      # Constraint: We cannot simply remove the `docker` key. The `lookup_from_k8s_field`
      # requires both `kubernetes` and `docker` keys to be present to perform enrichment.
```

dimiechatz (approved):
LGTM! The proposed fix makes sense. @polinashtanko thanks for the detailed explanation!

Let's roll it out in a test cluster to test it out before merging.

charts/lagoon-logging/templates/logs-dispatcher.fluent-conf.configmap.yaml
```diff
@@ -281,6 +281,23 @@ data:
...
+      <filter lagoon.*.router.nginx>
+        @type record_modifier
+        remove_keys _dummy_
+        <record>
+          _dummy_ ${record['docker']['container_id'] = "#{record.dig('kubernetes','namespace_name')}_nopod"; nil}
+        </record>
+      </filter>
```
anothertobi: Based on https://github.com/fluent-plugins-nursery/fluent-plugin-kubernetes_metadata_filter/blob/5aae4331454f7e386cdaa2022937f138611055c1/lib/fluent/plugin/filter_kubernetes_metadata.rb#L101, I assume that:

> Replacing the id with a fixed per-namespace string means at most ONE 404 API call per namespace; every subsequent record is a cache hit.

does not work because the 404 would still remain uncacheable. In any case, we can try this out in a test cluster and see what the results are.

psh396:
rolled out in test9 https://grafana.test6.amazee.io/goto/1S-gSPAvg?orgId=1

dimiechatz:
> rolled out in test9 https://grafana.test6.amazee.io/goto/1S-gSPAvg?orgId=1

I am not 100% sure if this solved the issue cause if we look back the last 30 days (https://grafana.test6.amazee.io/goto/yJUN8i1Dg?orgId=1), we can see peaks and valleys before rolling out the fix.

@polinashtanko if you could add a short explanation for what you also observe, it would be great.

PR #911, opened 2026-05-20
chore: change rollout strategy for keycloak

shreddedbacon:
Small change to the keycloak deployment strategy to be `Recreate` instead of the default `Rollout` to prevent issues with upgrading versions of keycloak.

This does mean that during updates, keycloak *will* be unavailable until the new pod is started.


bomoko (approved):
Make it so

PR #925, opened 2026-07-02
Exclude logging for a configurable list of namespaces (via regexp)

fubarhouse:
This PR will add a configurable list of regexp exclusions so that Fluentd can ignore any namespaces that match any of the configured regexp rules.

This is useful because, if we need to exclude a bunch of namespaces from logging (e.g., a particular client, project, or logging type), we can do so safely downstream of this chart without exposing information that might be sensitive.


chelios (approved):
Looks ok to me.

fubarhouse:
This has been successfully tested, so we're going to merge now.

PR #926, opened 2026-07-02
Allow exceptions to disabling namespace regexps

fubarhouse:
This PR will allow Fluentd configuration to exempt certain namespace regexp patterns from logging disablement, for when specific edge cases need to be configured on a case-by-case basis.


fubarhouse:
This has been tested under the following scenarios:

Exclude all namespaces under a path except for one subpath, and also except for two subpaths.
Having thought about this just now, I will run an additional test tomorrow morning.

PR #928, opened 2026-07-09
Insights-remote update

bomoko:
Bumps version of insights remote to v0.0.19


rocketeerbkw:
I included this in https://github.com/uselagoon/lagoon-charts/pull/929 since it bumps lagoon-remote charts version

PR #930, opened 2026-07-13
Exclude configured lagoon-logs-*

fubarhouse:
This MR will exclude lagoon-logs-* from the configured exclusion regexes (of namespaces) so that lagoon build logs will also not be included in logging collection based on regexp. So, if you need to enable a specific environment/project collection, there is still a good way to do that at a low scale.

Plans for a broader scale approach have been made and can be included if that becomes a legitimate use case.

Now ready for review, this has been tested thoroughly and validated on a test cluster.
Internally, this relates to PLAT-1270.

### DIFF
--- a/charts/lagoon-logging/Chart.yaml
+++ b/charts/lagoon-logging/Chart.yaml
@@ -19,7 +19,7 @@ type: application
 # time you make changes to the chart and its templates, including the app
 # version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.94.0
+version: 0.95.0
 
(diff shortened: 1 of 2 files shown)


dimiechatz:
@fubarhouse could you add some context in the PRs description and link any related AIOP/PLAT ticket? Also, has the change been tested in a test cluster?

fubarhouse:
@dimiechatz done - as this is a public repository, I have been intentionally vague about some details.

dimiechatz (approved):
LGTM! We'll need to update the chart version in helm catalog as well.

PR #941, opened 2026-08-17
feat: auto-generated hostkeys for lagoon-remote ssh-portal

shreddedbacon:
This enables automatically generated hostkeys for lagoon-remote ssh-portal. If ones are already defined in the values, they will be used. If automatically generated ones are created, they will also be re-used on updates.

Only `ed25519` will be generated by default, `rsa` and `ecdsa` are optional.


rocketeerbkw (approved):
Nice DX improvement to prevent duplicate host keys 🎉

PR #944, opened 2026-08-27
Add optional mTLS for CDN logs collector

britslampe:
The CDN logs collector can authenticate with a shared key defined in the fluentd.conf without custom TLS certificates and a custom CA. Certificates generated from a publicly trusted CA via a ClusterIssuer removes the need for the custom CA certificate but keeps TLS transport in place.

### DIFF
--- a/charts/lagoon-logging/values.yaml
+++ b/charts/lagoon-logging/values.yaml
@@ -175,6 +175,9 @@ cdnLogsCollector:
   # If TLS configuration not required disable mTLS.
   # mTLS: false
 
+  # Required when mTLS is disabled.
+  authToken: ""
+
   # TLS configuration is required
   # These should be server certificates, and the CDN should be configured to
   # present client certificates with the same trust root.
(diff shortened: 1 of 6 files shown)


charts/lagoon-logging/values.yaml
```diff
@@ -172,6 +172,9 @@ cdnLogsCollector:
     type: LoadBalancer
     annotations: {}
 
+  # If TLS configuration not required disable mTLS.
+  # mTLS: false
```
fubarhouse: This needs to be uncommented to be used as the default value

fubarhouse:
This looks pretty coherent and sensible.

Please let me know either here or privately the outcome of your tests, and I'll be happy to approve it.

fubarhouse:
Rebased, updated and tagged so that tests are run.

Based on the outcome of the questions regarding `authToken`, I'm happy to approve.
Send me a DM after you've resolved that thread.

britslampe:
@fubarhouse got some feedback on the logic I used for the default, it's now flipped so that the original architecture is retained. The values.yaml uses `mTLS: true`.

rocketeerbkw (approved):
Discussed out of band and LGTM.

PR #946, opened 2026-09-08
Field exclusion for traefik router logs

fubarhouse:
Adds support for removing Traefik router logs that match a set of regular expressions, configurable per implementation.


chelios (approved):
Looks good and confirmed in test6

fubarhouse:
As described above, A/B testing has passed on a test cluster.

